PR #7, opened 2020-03-06
Mock testing and extending implementations

sblack-usu:
- Added a mocking framework with tests that have full coverage over all methods except for some response parsing that I was unable to find examples for
- Fixed pmcodes
- introduced metadata.  Each method returns the DataFrame along with the metadata as a tuple
- implemented water use query
- implemented ratings query
- fixed results query
- fixed and implemented sites/what_sites queries
- removed some code that had no references
- did some reorganizing to make the code easier to maintain

I also added a python notebook that attempts to recreate the R dataretrieval vignette, I used this to guide most of my work.  The vignette doesn't provide examples for all functionality so I will need some more direction on what functionality is missing.  

My known TODOs are:
- implement rename
- implement reshape
- add metadata for statistic_info and disclaimer (there may be other metadata fields I haven't found yet)
- All functions that don't have examples in the vignette


dataretrieval/wqp.py
```diff
@@ -57,10 +57,10 @@ def get_results(**kwargs):
     kwargs['zip'] = 'no'
     kwargs['mimeType'] = 'csv'
 
-    response = query(wqp_url('Result'), **kwargs)
+    response = query(wqp_url('Result'), list(kwargs.items()))
```
thodson-usgs, line 60: Why change **kwargs to list(kwargs.items())?

thodson-usgs:
I'd prefer get_data functions to return a single dataframe, so that we don't break existing code.  BUT I also realize there's no great way to replicate how RdataRetrieval stores metadata. Another option is to maintain the original default behavior but add a kwarg to return additional metadata. Thoughts?

dataretrieval/wqp.py
sblack-usu: query parameters can have multiple keys with different values.  kwargs is a dictionary, which only can have one value per key.  This is why get_pmcodes was not working, the show parameter has 4 parameters with the keyword show.  Since it was getting placed into a dictionary, only the last value was being used.

sblack-usu:
> 
> 
> I'd prefer get_data functions to return a single dataframe, so that we don't break existing code. BUT I also realize there's no great way to replicate how RdataRetrieval stores metadata. Another option is to keep the original behavior but add a kwarg to return additional metadata. Thoughts?
> 
> IF we change the default behavior, then update examples in README to match.

Totally understand the concern here.  I've waffled quite a bit on how to handle the metadata.  This is the most pythonic way I could come up with for handling the metadata.  If you look at older commits, I actually started with monkey patching (very similar to what the R library is doing) which returns the DataFrame with extra fields.  i felt it was too implicit though.

I considered creating a class that wraps the dataframe and metadata, but that breaks existing code as well and requires one more step to get the dataframe.  My understanding is that people are more interested in the dataframe than the metadata so I nixed that idea.

It's incredibly simple to just ignore the metadata as this PR shows so it wouldn't be hard to uplift existing code.  We can also consider what you're suggesting with a flag for the metadata.  Another possibility is a different method entirely, which I may consider if I can setup some better patterns across all apis for the client.  Yet another option is to extend the DataFrame class and add the methods/fields we want.  This would be very similar to monkey patching but it would be explicit.

thodson-usgs:
I'll pose two options for how to resolve metadata, otherwise I'm okay with everything else proposed in this pull request.
1. Have a kwarg that switches whether *get_data* returns just a dataframe or a dataframe with a metadata object. Or/and return only a dataframe and store the associated metadata in the experimental Dataframe.attrs attribute. I *suspect* the lack of metadata support in Pandas Dataframes is a common complaint, and that there will be better support in the future. Rather than change the API now to add metadata objects, then change it back again once metadata is incorporated into Pandas, I'd prefer to keep the API static and instead include metadata as an experimental feature using Dataframe.attrs. 
2. Although we prefer to keep API's static, there are good reasons to change them now and then. If you believe this is an instance, let's discuss it further on Friday's telecon.

sblack-usu:
I'd like to avoid DataFrame.attrs, it's experimental and I don't think it really buys us much.  There isn't much information about it and it's subject to change.

by _get_data_ I think you are referring to _get_record_.  We could just leave _get_record_ without metadata and implement a new interface that includes metadata.  I don't think _get_record_ is very descriptive or any easier to use than the individual methods.  I'm also searching for ways to implement something similar to _get_record_ that could be easier to work with and document parameters better while also normalizing some of the parameters.  Lots to talk about tomorrow.

thodson-usgs:
Closing the pull request based on our phone call. Get_record is my own convenience function, which has no equivalent in the R package. We can deprecate it , but I'd prefer to keep its behavior static, at least for now. Otherwise, I'm flexible about creating new functions or changing existing functions.

sblack-usu:
sounds good, I'll make some of the changes we talked about and preserve get_record.  I usually work on this package at the end of each week.

PR #15, opened 2020-07-04
Raise a custom error if the RDB is an error report

elmerehbi:


thodson-usgs:
Sorry for the delay. 
Do you recall which function generated the error? dataretrieval pulls from a few different data portals. As is, your patch will only work for one of them, but it shouldn't be hard to fix.

elmerehbi:
Hello. 
I think I was using  `nwis.get_record` to retrieve `iv` values.

elmerehbi:
Bumping this. Otherwise we can close it.

PR #18, opened 2020-11-12
Create python-package.yml

thodson-usgs:


thodson-usgs:
@sblack-usu 
I'm trying to setup a github action to run tests then update pipy whenever someone pushes to master.
But the mock requests are not getting parsed correctly during the tests:
FAILED waterservices_test.py::test_get_iv - AttributeError: 'Index' object has no attribute 'tzinfo' 

Would you have a look? I'm not familiar with requests-mock.

sblack-usu:
yeah, I'll take a look.

sblack-usu:
I just did a quick run of the test without mocking and got the same error. 
```
from dataretrieval.nwis import get_iv
df, md = get_iv(sites=["01491000", "01645000"], start='2019-02-14', end='2020-02-15')
```
results with
```
"dataretrieval/nwis.py", line 38, in format_response
    if df.index.levels[1].tzinfo is None:
AttributeError: 'Index' object has no attribute 'tzinfo'
```

This means it's not an issue with the mock, rather the parsing isn't working anymore.  I saw a PR go through earlier that had to do with tzinfo, I think that's a good place to start.  I'll take a look and see if anything stands out to me.

thodson-usgs:
It's site specific. The following works for me.

```
df, md = get_iv(sites='03339000', start='2020-11-1')
```

sblack-usu:
It might not be site specific.  The code is different when there is one site than when there is more than one site.
https://github.com/USGS-python/dataretrieval/blob/master/dataretrieval/nwis.py#L38

Is there a reason it's treated differently?

thodson-usgs:
This also works.
```
df, md = nwis.get_iv(sites=['05586300','03339000'], start='2020-11-1')
```
The main difference is how the dataframes get indexed. Multi-site queries get an multi-level index (time and site), whereas single site queries are indexed by time only.

I'm not opposed to changing that behavior, but let's table that.

sblack-usu:
Ok, then it seems like we should look for the tzinfo attribute before trying to access it.  Something like this.
```
    elif len(df['site_no'].unique()) > 1:
        # setup multi-index
        df.set_index(['site_no', 'datetime'], inplace=True)
        if hasattr(df.index.levels[1], 'tzinfo') and df.index.levels[1].tzinfo is None:
            df = df.tz_localize('UTC', level=1)

    else:
        df.set_index(['datetime'], inplace=True)
        if hasattr(df.index, 'tzinfo') and df.index.tzinfo is None:
            df = df.tz_localize('UTC')
```

tests/waterservices_test.py
```diff
@@ -82,7 +82,7 @@ def test_get_iv(requests_mock):
     site = '01491000%2C01645000'
     request_url = 'https://waterservices.usgs.gov/nwis/iv?format={}' \
                   '&startDT=2019-02-14&endDT=2020-02-15&sites={}'.format(format, site)
-    response_file_path = 'data/waterservices_iv.txt'
+    response_file_path = 'tests/data/waterservices_iv.txt'
```
sblack-usu: Why is `tests/` put before each response_file_path?  That path doesn't exist relative to this file and all the tests will fail.

thodson-usgs: That's for the github action, which runs pytest from the dataretrieval directory by default. Not sure how to change that behavior.

thodson-usgs: I suppose one would cd to 'tests' during the action. Is that your preference?

sblack-usu: I would prefer the change directory.  That'll allow us to run tests individually from an IDE.

sblack-usu: Actually, I can change the directory it runs from in my IDE.  I'm fine either way, whichever you think is best.

PR #55, opened 2022-11-21
Add inplace Argument to dropna

douglasdennis:
While I was reviewing the code for something else I noticed a `dropna` call that didn't have `inplace` set to True and wasn't getting assigned to a new variable. I set inplace to True as it seemed that was what was being done elsewhere in the code. If you prefer the dropna to be removed then I can make that edit instead.

### DIFF
--- a/dataretrieval/nwis.py
+++ b/dataretrieval/nwis.py
@@ -58,7 +58,7 @@ def format_response(df, service=None, **kwargs):
 
 def preformat_peaks_response(df):
     df['datetime'] = pd.to_datetime(df.pop('peak_dt'), errors='coerce')
-    df.dropna(subset=['datetime'])
+    df.dropna(subset=['datetime'], inplace=True)
     return df
 
 
--- a/tests/nwis_test.py
+++ b/tests/nwis_test.py
@@ -1,5 +1,7 @@
+import numpy as np
+import pandas as pd
 import pytest
-from dataretrieval.nwis import get_record
+from dataretrieval.nwis import get_record, preformat_peaks_response
 
 START_DATE = '2018-01-24'
 END_DATE   = '2018-01-25'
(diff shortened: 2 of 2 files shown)


elbeejay:
Nice catch @douglasdennis and thanks for opening this PR. Would you mind adding a unit test that ensures we get the right behavior moving forward? 

Something relatively small like the following should work, you could add it to `tests/nwis_test.py`. It fails for me using the current `master` branch, but will pass once your change is implemented.

```
def test_iv_service_answer():
    df = test_iv_service()

    # check multiindex function
    assert df.index.names == [SITENO_COL, DATETIME_COL], "iv service returned incorrect index: {}".format(df.index.names)


if __name__=='__main__':
     test_measurements_service_answer()
     test_iv_service_answer()


def test_preformat_peaks_response():
    # make a data frame with a "peak_dt" datetime column
    # it will have some nan and none values
    data = {"peak_dt": ["2000-03-22",
                        "2001-06-08",
                        np.nan,
                        "2002-04-29",
                        "2003-02-23",
                        None,
                        "2003-12-12"],
            "peak_va": [1000,
                        2000,
                        3000,
                        4000,
                        5000,
                        6000,
                        7000]
    }
    # turn data into dataframe
    df = pd.DataFrame(data)
    # run preformat function
    df = preformat_peaks_response(df)
    # assertions
    assert 'datetime' in df.columns
    assert np.isnan(df['datetime']).sum() == 0
```

douglasdennis:
I'd be happy to add a test. I should be able to get that done tomorrow. Thank you for putting one together for me!

douglasdennis:
@elbeejay Sorry for the delay but I added the unit test. The test failure on the previous run looks to be an issue with setting up the build job and not related to this PR. Let me know if I'm missing something though.

elbeejay (approved):
This looks great, thanks @douglasdennis! Feel free to reach out in the future if you run into problems using `dataretrieval` or need help designing queries or manipulating the data.

PR #56, opened 2022-11-22
warnings for qwdata deprecation

elbeejay:
- adds deprecation warnings to calls to the qwdata service as well as a warning in the documentation
- drops CI support for python 3.6 as it is no longer included in the "ubuntu-latest" image (hence the failing CI build)

@lstanish-usgs this is the extent of where we would place warnings for this deprecation - feel free to propose changes to the wording.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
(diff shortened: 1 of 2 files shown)


elbeejay:
@thodson-usgs let me know what you think. Very minor PR but would rather not merge PRs without someone else laying eyes on 'em.

thodson-usgs (commented):
@lstanish-usgs could we give a date for which this retirement is planned? Or else ask Laura to revise the vignette?
I realize it could change, but the vignette says the web service will shut down in spring 2022, which is confusing.

lstanish-usgs:
@thodson-usgs Thanks for catching that. I think we should update the vignette, will reach out to Laura.

Issue #61, opened 2022-12-13
Better documentation of time zones and time zone awareness of objects would be helpful


ddileonardo:
It is unclear to a new user how the time zone is represented in the retrieved data. This is in large part due the a lack of documentation of the pytz module (at least what I could find). From what I can tell the time information is retrieved in the following way:

- Times are given as time zone aware with the time zone indicated by pytz.FixedOffset from UTC with the amount of the fixed offset representing minutes (i.e. pytz.FixedOffset(300) is 300 minutes = 5 hours)
- Time zones are also listed in the site info
- Example time stamp retrieved using get_record: time = Timestamp('2022-10-01 00:00:00-0500', tz='pytz.FixedOffset(-300)')
- Time zones can be converted using the pandas conversion tools
- For the example, time. tz_convert(tz=None), removes the time zone information and returns a UTC time --> Timestamp('2022-10-01 05:00:00')
- time.tz_convert(tz='utc') yields the same UTC time, but with the time zone information still attached --> Timestamp('2022-10-01 05:00:00+0000', tz='UTC')
- time.tz_convert(tz='America/New_York') converts the time to a new time zone --> Timestamp('2022-10-01 01:00:00-0400', tz='America/New_York')

If this is correct and helpful, please feel free to use in the documentation.

elbeejay:
This is a great point @ddileonardo, I'll review the individual bullet points as I address issue #60 and put together PR #62 to make sure it is more clear how time zones are being handled and what the output means.

Issue #60, opened 2022-12-13
Error when combining `Dataframe` from stations with different time zones


SorooshMani-NOAA:
When we try to get bulk station data (e.g. instantaneous values for a given state), if some stations have different time zones, `pandas` runs into issue when trying to merge station data frames.

To reproduce:
```python
import dataretrieval.nwis as nwis

sites, sites_md = nwis.what_sites(stateCd='MD')
 iv, iv_md = nwis.get_iv(sites=sites.site_no.values[:25].tolist())
```

The results:
```
.
.
.
File ~/miniconda3/envs/data/lib/python3.10/site-packages/pandas/core/reshape/merge.py:1349, in _MergeOperation._maybe_coerce_merge_keys(self)
   1347     raise ValueError(msg)
   1348 elif not needs_i8_conversion(lk.dtype) and needs_i8_conversion(rk.dtype):
-> 1349     raise ValueError(msg)
   1350 elif isinstance(lk.dtype, DatetimeTZDtype) and not isinstance(
   1351     rk.dtype, DatetimeTZDtype
   1352 ):
   1353     raise ValueError(msg)

ValueError: You are trying to merge on object and datetime64[ns, pytz.FixedOffset(-300)] columns. If you wish to proceed you should use pd.concat

```

Information for each iteration of merging new station:
```
--> 0   2022-12-13 09:45:00-05:00
Name: dateTime, dtype: datetime64[ns, pytz.FixedOffset(-300)]
==> datetime64[ns, pytz.FixedOffset(-300)]
--> 0   2022-12-13 09:45:00-05:00
Name: dateTime, dtype: datetime64[ns, pytz.FixedOffset(-300)]
==> datetime64[ns, pytz.FixedOffset(-300)]
--> 0   2022-12-13 09:48:00-05:00
Name: dateTime, dtype: datetime64[ns, pytz.FixedOffset(-300)]
==> datetime64[ns, pytz.FixedOffset(-300)]
--> 0   2009-10-01 00:45:00-04:00
Name: dateTime, dtype: datetime64[ns, pytz.FixedOffset(-240)]
[...]

PR #62, opened 2022-12-13
Error when merging across timezones

elbeejay:
PR to close issue #60 

Issue is well described in #60. A test has been added that will be expanded in the future when functionality is fixed.

Fix is to defer setting the time objects when calling `pd.read_json()` by setting the parameter `convert_dates=False` in that function call. Then once the final data frame is constructed, the datetime information is converted from a string to a `datetime.datetime` object. So the end result should be the same as it was before, just the internal conversion happens in a different order. I wrote a pair of tests based on the error described in the issue to confirm the original behavior / typing using sites for which the old code worked as well as a test where it failed. Then with the proposed changes both tests pass. 

PR should also close issue #61 

Points raised around time handling there prompted the creation of a "User Guide" section in the documentation where little bits of code and explanation can be added to discuss things like the timestamps and how `dataretrieval` attempts to handle datetimes. 

### Change in UTC localization

As a result of some of the poking around related to #61, I found that while `dataretrieval` attempts to UTC-localize the timestamps (in `nwis.format_response`), this often wasn't happening for the multi-site case. This resulted in some of the timestamp stuff discussed in #61, such as the `pytz.FixedOffset` portions of the timestamps. Since this PR more manually converts the timestamps now (rather than letting `pd.read_json()` try to do it), we gained the opportunity to set the times in UTC at that point. So now even the multi-site returns should have timestamps that are formatted in UTC by default. This seems like it was the intent of the package all along, but just wasn't always happening.


elbeejay:
This PR should be good to go, the failing test involves a service that wasn't modified as part of this PR, so it may just be down temporarily. Aim of this PR is to address issues #60 and #61, so @SorooshMani-NOAA and @ddileonardo please feel free to check out the PR description and code and I welcome any feedback you may have.

SorooshMani-NOAA:
@elbeejay it has fixed the original issue. Maybe unrelated, but when I try to get the data from  **all** the sites in Maryland, it gives me an error related to parsing the JSON. My assumption is maybe the `site` is too large and the webAPI responds with an error (?) which fails to be parsed. Is there any way to have better error messages? e.g. area is too large for fetching data?

elbeejay:
> @elbeejay it has fixed the original issue. Maybe unrelated, but when I try to get the data from **all** the sites in Maryland, it gives me an error related to parsing the JSON. My assumption is maybe the `site` is too large and the webAPI responds with an error (?) which fails to be parsed. Is there any way to have better error messages? e.g. area is too large for fetching data?

Glad to hear the original issue is handled by this fix. The new issue you've run into when trying to fetch data for all sites in Maryland is that the query size is too large; in #65 a more informative error message is proposed (as the current one is, as you say, not very informative).

thodson-usgs (approved):

Issue #66, opened 2022-12-20
Discrepancy between GitHub tag and PyPI version


SorooshMani-NOAA:
The GitHub repository is tagged as `v1.0.0` however the version on PyPI is still at `0.8`:

https://github.com/USGS-python/dataretrieval/releases/tag/v1.0.0
vs
https://github.com/USGS-python/dataretrieval/blob/46eb1dd08ef96901644658a85fa88735c0c387e4/setup.py#L3-L4

elbeejay:
@thodson-usgs released v1.0.1 and it deployed to pypi so release now matches pypi.

thodson-usgs:
Forgot to close this in the commit.

PR #67, opened 2022-12-20
Fix to enable seriesCatalogOutput

### LINKED ISSUE #34: Parameter code not returned from site information
Hi, I am trying to retrieve the site information by given a huc region, but the `parm_cd` is not returned as one of the columns. I have used R-dataretrieval in the past, and it has always returned the `parm_cd` for me. I wonder if similar information could be returned using this python package.

This is what I have used in the R package:

```
df <- readNWISdata(huc = HUCcode, parameterCd = pCode, service="site", 
                     seriesCatalogOutput=TRUE)
```

The dataframe returned the following columns:
```
'agency_cd''site_no''station_nm''site_tp_cd''dec_lat_va''dec_long_va''coord_acy_cd''dec_coord_datum_cd''alt_va''alt_acy_va''alt_datum_cd''huc_cd''data_type_cd''parm_cd''stat_cd''ts_id''loc_web_ds''medium_grp_cd''parm_grp_cd''srs_id''access_cd''begin_date''end_date''count_nu'
```

This is what I used in python:
```
df = nwis.get_record(huc = HUCcode, parameterCd= pCode, service='site',
                   seriesCatalogOutput = True)
```
And this is what's returned:
```
'agency_cd', 'site_no', 'station_nm', 'site_tp_cd', 'lat_va', 'long_va',
       'dec_lat_va', 'dec_long_va', 'coord_meth_cd', 'coord_acy_cd',
       'coord_datum_cd', 'dec_coord_datum_cd', 'district_cd', 'state_cd',
       'county_cd', 'country_cd', 'land_net_ds', 'map_nm', 'map_scale_fc',
       'alt_va', 'alt_meth_cd', 'alt_acy_va', 'alt_datum_cd', 'huc_cd',
       'basin_cd', 'topo_cd', 'instruments_cd', 'construction_dt',
       'inventory_dt', 'drain_area_va', 'contrib_drain_area_va', 'tz_cd',
       'local_time_fg', 'reliability_cd', 'gw_file_cd', 'nat_aqfr_cd',
       'aqfr_cd', 'aqfr_type_cd', 'well_depth_va', 'hole_depth_va',
       'depth_src_cd', 'project_no'
```

The columns are also missing `begin_date`, `end_date`, and `count_nu`, which would be very helpful.

Any idea?

elbeejay:
This PR closes #34.

Now in `get_info` the `kwargs` are checked for the `seriesCatalogOutput` parameter which, if true, is set in `kwargs` and the expanded rdb output is not requested. If `seriesCatalogOutput` is false or not present, then the expanded rdb output is requested -- so the default behavior of this function does not change.

Tests and an example in the documentation are added to show how this argument works.

Additional details surrounding the original problem are described in #34. 

A section in the documentation is added on updating the version numbers, which was prompted by #66 which still needs to be resolved.

### DIFF
--- a/dataretrieval/nwis.py
+++ b/dataretrieval/nwis.py
@@ -455,17 +455,23 @@ def get_info(**kwargs):
         siteOutput=expanded cannot be used if seriesCatalogOutput=true or with
         any values for outputDataTypeCd.
 
-    seriesCatalogOutput : boolean
+    seriesCatalogOutput : boolean or string
         A switch that provides detailed period of record information for
         certain output formats. The period of record indicates date ranges for
         a certain kind of information about a site, for example the start and
-        end dates for a site's daily mean streamflow.
+        end dates for a site's daily mean streamflow. Can be set as True,
+        'True', 'TRUE', or 'true'.
 
     For additional parameter options see
     https://waterservices.usgs.gov/rest/Site-Service.html#stateCd
     """
-
-    kwargs['siteOutput'] = 'Expanded'
+    seriesCatalogOutput = kwargs.pop('seriesCatalogOutput', None)
+    if seriesCatalogOutput in ['True', 'TRUE', 'true', True]:
+        # convert bool to string if necessary
+        kwargs['seriesCatalogOutput'] = 'True'
+    else:
+        # cannot have both seriesCatalogOutput and the expanded format
+        kwargs['siteOutput'] = 'Expanded'
 
     response = query_waterservices('site', **kwargs)
 
--- a/dataretrieval/nwis.py
+++ b/dataretrieval/nwis.py
@@ -455,12 +455,11 @@ def get_info(**kwargs):
         siteOutput=expanded cannot be used if seriesCatalogOutput=true or with
         any values for outputDataTypeCd.
 
-    seriesCatalogOutput : boolean or string
+    seriesCatalogOutput : boolean
         A switch that provides detailed period of record information for
         certain output formats. The period of record indicates date ranges for
         a certain kind of information about a site, for example the start and
-        end dates for a site's daily mean streamflow. Can be set as True,
-        'True', 'TRUE', or 'true'.
+        end dates for a site's daily mean streamflow.
 
     For additional parameter options see
     https://waterservices.usgs.gov/rest/Site-Service.html#stateCd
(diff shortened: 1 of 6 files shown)


dataretrieval/nwis.py
```diff
@@ -455,17 +455,23 @@ def get_info(**kwargs):
...
-        end dates for a site's daily mean streamflow.
+        end dates for a site's daily mean streamflow. Can be set as True,
+        'True', 'TRUE', or 'true'.
 
     For additional parameter options see
     https://waterservices.usgs.gov/rest/Site-Service.html#stateCd
     """
-
-    kwargs['siteOutput'] = 'Expanded'
+    seriesCatalogOutput = kwargs.pop('seriesCatalogOutput', None)
```
thodson-usgs, line 467: Fine, but also consider:
`kwargs.get('seriesCatalogOutput')

dataretrieval/nwis.py
```diff
@@ -455,17 +455,23 @@ def get_info(**kwargs):
         siteOutput=expanded cannot be used if seriesCatalogOutput=true or with
         any values for outputDataTypeCd.
 
-    seriesCatalogOutput : boolean
+    seriesCatalogOutput : boolean or string
```
thodson-usgs: Better to stick with boolean and convert to string behind the scenes, don't you think?

thodson-usgs (changes requested):
I've made two suggestions. Neither is required.

dataretrieval/nwis.py
elbeejay: Yes, I agree with this, it is more clear if we document this as expecting a boolean `True` value.

dataretrieval/nwis.py
elbeejay: In my view the benefit of using `pop` to explicitly remove whatever value is present for `seriesCatalogOutput` from the dictionary is that any value that is not translated to `'True'` is not added as part of the API query string.

I think any values other than `'True'` are ultimately ignored by the web service, but I also see no reason to allow any arbitrary string to be injected via the `seriesCatalogOutput` parameter. In the extreme someone could pass something like `seriesCatalogOutput='False&site_no=123456'` and the final API query might work. Using `pop` allows us to explicitly "sanitize" the `seriesCatalogOutput` parameter and we should be confident in the future that it won't be the cause of some sort of API response error.

PR #72, opened 2022-12-22
Improve documentation

### LINKED ISSUE #33: Which data portals does this package pull from?
Readme states that it can pull from "a variety of portals" -- would be great if those were listed and updated!  Thanks!

### LINKED ISSUE #46: Dataframe not returned from get_water_use()
Today I tried using the get_water_use() function

```
from dataretrieval import nwis

wudict = {'AZ':['Mohave', 'Graham', 'Cochise', 'Pinal', 'Gila', 
                'Maricopa', 'Yavapai'],
          'NM':['Catron', 'Grant'],
          'UT':['Washington']}

for state in wudict.keys():
    counties = wudict[state]    
    wu = nwis.get_water_use(years='ALL', state=state, counties=counties, 
                            categories='ALL')
    wu_df = wu[0]
```

and the return did not produce a dataframe, instead I got something that was coerced to a dataframe (wu_df) but is definitely not the return I was looking for (see image below for picture of what was returned). 

image

What's going on here? I tried updating the package, but that didn't solve it.

thanks,
Annie

### LINKED ISSUE #68: Functionality or example to handle large requests
I think that the problem encountered in #64 is likely somewhat common.

There are 2 ways (in my view) that we could help users make larger queries:

1. Create functionality to split up or "chunk" large requests and submit separate queries either in series or parallel to the web service and then return one large merged data frame. This would be handy for the user, but at the cost of increasing the complexity of what is intended to be a fairly thin and simple package.

2. We can create a few examples of how one might write some code to do the "chunking" of large queries themselves, and wrap existing ``dataretrieval`` functions within a for-loop or similar. This documentation could be referenced in the 414 error message we return, and would not make the package itself any more complex. This option keeps the package light-weight and simple by offloading the complexity associated with large queries to the user.

Thoughts?

elbeejay:
## Improvements to documentation and doc-strings

This looks like a large PR, but it mostly consists of cosmetic changes aimed at improving the rendered documentation and standardizing some of the doc-string and code formatting throughout the package. Given the large size, it might be best to spread the review out over a few days (or weeks) and go through changed files individually due to the number of changes. These changes are summarized below:

### Closing Issues

- Closes #68 by improving the quality of the error message associated with the 414 error message to include a pseudo-code example of splitting up a large query into chunks of some size `n`
- Closes #33 by creating a new page in the documentation with a table of the various data portals ``dataretrieval`` pulls data from and their corresponding URLs
- Closes #46 by adding an example of collecting water use data for a single county in Arizona to the doc-string

### Additional Examples

- Adds at least 1 if not more examples of function usage to the doc-strings of public functions. These will render below the API documentation for a given function, ex:

image

- Adds (and dynamically renders) the hydroshare notebooks put together by @horsburgh - notebooks were downloaded from https://www.hydroshare.org/resource/c97c32ecf59b4dff90ef013030c54264/ - they do not directly reference the notebooks but reference the downloaded copies in the `demos/` subdirectory, it might be possible to pull the latest versions of the notebooks with some fancy CI stuff to download and extract the source notebooks to the proper location every time the documentation workflow is run

### Doc-strings and linting

- PR includes a bunch of edits to standardize the doc-strings to the numpy format. This includes ensuring headers are `Parameters`, `Other Parameters`, `Returns`, and `Examples` exclusively in the doc-strings, as well as explicit inclusion of parameter types
- Some linting to conform to standard PEP-standards for spacing in arguments, line lengths, etc.
- Refer to Python objects in the documentation with double backquotes for Python functions and objects outside of the language, e.g., ``pandas.DataFrame`` whereas for functions that are within the package, the convention :obj:`object` is used, e.g., :obj:`dataretrieval.nwis.get_record` as Sphinx is able to automatically link to the relevant API documentation in that case
- Execute much of the documented code, anything preceded by `.. doctest::` will be run when the documentation is built, some code is preceded by `.. code::` instead because that code is either costly to run (and therefore impractical for the CI job) or requires `gdal` which is not installed on the CI runner at this time

### Code Changes

- Adds a private function `_alter_kwargs` to `wqp.py` to reduce code repetition by using a function to handle the keyword arguments that are not currently supported by `dataretrieval`
- Changes the URLs and paths for `nadp.py` to reflect the current addresses and URL-conventions for the national trends and mercury deposition maps and data

### Unit Tests

- Adds tests for the NADP functions with a `nadp_test.py` file. These tests are marked to `xfail` as the CI runner doesn't have `gdal` installed, but they should pass locally when running the test suite
- Adds unit tests for the new `_alter_kwargs` function within the `wqp_test.py` file

### DIFF
--- a/dataretrieval/utils.py
+++ b/dataretrieval/utils.py
@@ -11,9 +11,32 @@ from dataretrieval.codes import tz
 def to_str(listlike, delimiter=','):
     """Translates list-like objects into strings.
 
+    Parameters
+    ----------
+    listlike: list-like object
+        An object that is a list, or list-like
+        (e.g., :obj:`pandas.core.series.Series`)
+
+    delimiter: string, optional
+        The delimiter that is placed between entries in listlike when it is
+        turned into a string. Default value is a comma.
+
     Returns
     -------
-        List-like object as string
+    listlike: string
+        The listlike object as string separated by the delimiter
+
+    Examples
+    --------
+
+    .. doctest::
+
+        >>> to_str([1, 'a', 2])
+            '1,a,2'
+
+        >>> to_str([0, 10, 42], delimiter='+')
+            '0+10+42'
+
     """
     if type(listlike) == list:
         return delimiter.join([str(x) for x in listlike])
@@ -119,6 +142,8 @@ def update_merge(left, right, na_only=False, on=None, **kwargs):
 
 
 class Metadata:
+    """Custom class for metadata.
+    """
     url = None
     query_time = None
     site_info = None
@@ -133,6 +158,8 @@ class Metadata:
 
 
 def set_metadata(response):
+    """Function to initialize and set metadata from an API response.
+    """
     md = Metadata()
     md.url = response.url
     md.query_time = response.elapsed
@@ -148,13 +175,16 @@ def query(url, payload, delimiter=','):
 
     Parameters
     ----------
-        url :
-        payload : query parameters passed to requests.get
-        delimiter : delimeter to use with lists
+        url: string
+            URL to query
+        payload: dict
+            query parameters passed to requests.get
+        delimiter: string
+            delimeter to use with lists
 
     Returns
     -------
-        string : query response
+        string: query response
     """
 
     for key, value in payload.items():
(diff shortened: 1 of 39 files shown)


dataretrieval/utils.py
```diff
@@ -119,6 +136,8 @@ def update_merge(left, right, na_only=False, on=None, **kwargs):
 
 
 class Metadata:
```
thodson-usgs, line 138: This class isn't idiomatic. Note to self to fix.

thodson-usgs (approved):
Excellent!

PR #74, opened 2022-12-23
Expand site info metadata (addresses #73)

### LINKED ISSUE #73: `site_info` metadata for queries with `stateCd`, `huc`, or `bBox`
Currently only metadata associated with queries that have `sites` or `site_no` has a valid `site_info` function. Is there any reason that this is not extended to all types of queries? e.g. by state code, by HUC or even bounding box?

https://github.com/USGS-python/dataretrieval/blob/2c8ed7d773084e4d426ad8b89ee1510d036393b9/dataretrieval/nwis.py#L847-L852

elbeejay:
This PR closes #73 and fixes bug with `countyCd` queries.

---

Specifically the following changes are proposed:

- Allows water service queries (`query_waterservices`) to be called with only the `countyCd` being specified, previously this was not allowed and made it impossible? to do a county query. Alters error message to match expanded behavior.
- Expands the `_set_metadata` function to allow the `lambda` function assigned to the `site_info` attribute of the metadata object to be set by parameters other than only `sites` and `site_no`
- Adds tests of this expanded `_set_metadata` functionality that do not execute any queries but check that the `lambda` functions behind `metadata.site_info` are set up
- Relaxes metadata check for `site_info` in `waterservices_test.py` as the `site_info` attribute can now be set by more parameters

### DIFF
--- a/dataretrieval/nwis.py
+++ b/dataretrieval/nwis.py
@@ -335,8 +335,8 @@ def query_waterservices(service, **kwargs):
     Usage: must specify one major filter: sites, stateCd, bBox,
 
     """
-    if not any(key in kwargs for key in ['sites', 'stateCd', 'bBox', 'huc']):
-        raise TypeError('Query must specify a major filter: sites, stateCd, bBox, or huc')
+    if not any(key in kwargs for key in ['sites', 'stateCd', 'bBox', 'huc', 'countyCd']):
+        raise TypeError('Query must specify a major filter: sites, stateCd, bBox, huc, or countyCd')
 
     if service not in WATERSERVICES_SERVICES:
         raise TypeError('Service not recognized')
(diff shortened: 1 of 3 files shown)


dataretrieval/nwis.py
```diff
@@ -845,12 +845,24 @@ def _set_metadata(response, **parameters):
             unpacked dictionary of the parameters supplied in the request
     """
     md = set_md(response)
-    site_aliases = ['sites', 'site_no']
-    for alias in site_aliases:
-        if alias in parameters:
-            md.site_info = lambda: what_sites(sites=parameters[alias])
-            break
+    # site_no is preferred over sites to set site_info if both are present,
```
thodson-usgs, line 848: This part feels a little kludgy, but I'd have to think more about how we improve it.

thodson-usgs (approved):

PR #81, opened 2023-01-31
Revised Contributing File

### LINKED ISSUE #80: Update CONTRIBUTING.md
CONTRIBUTING.md is showing its age. We should revise it with something boilerplate that could be adopted as the standard for other USGS projects.

elbeejay:
I'd be receptive to feedback on these proposed changes to the `CONTRIBUTING.md` file. I don't think there is going to be a "one-size-fits-all" approach for these contributing guidelines as some of the expectations, standards, and syntax will vary between projects and programming languages.

intended to close #80

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,10 @@
 # Contributing
 
-Contributions to dataretrieval are welcome and greatly appreciated, but please
-read this document before doing so.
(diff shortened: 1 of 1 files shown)


thodson-usgs (approved):
@elbeejay, great work! This is a comprehensive template that developers can modify to suite their projects.

PR #85, opened 2023-02-10
Correct the path to _version.py in pyproject.toml

### LINKED ISSUE #84: Typo in project.toml
Don't recall where we'd decided version information would live, but either way the current reference in `project.toml` looks wrong.

[Line 44](https://github.com/USGS-python/dataretrieval/blob/ba4ddef7e0476b3aac9d65307c605ee59863a6a9/pyproject.toml#L44):

```
write_to = "ratingcurve/_version.py"
```

thodson-usgs:


elbeejay (changes requested):
Need to add a `_version.py` file (at least I don't see one in this repository)

thodson-usgs:
I just learned about this the hard way, when I accidentally committed _version.py on another project. Don't worry it's been added to .gitignore.
_version.py is dynamically generated by setuptools. Basically, setuptools will pull the version info from the most recent tag, apply that version to the PyPI upload and write a new _version.py, so that dataretrieval.__version__ returns the correct version info.

elbeejay (approved):

PR #86, opened 2023-02-10
Defining a user-agent in GET requests

### LINKED ISSUE #83: Add a user agent for GET requests
Add a "user agent" to our GET requests that adds embeds information about the query such as the fact that it is coming from python dataretrieval package version xx.xx 

This is something that is implemented in R: [function](https://github.com/DOI-USGS/dataRetrieval/blob/e57ce78eab1b0dcd02c3604a1958c2454ab0452e/R/getWebServiceData.R#L100), use in [GET request](https://github.com/DOI-USGS/dataRetrieval/blob/e57ce78eab1b0dcd02c3604a1958c2454ab0452e/R/getWebServiceData.R#L148)

Can help establish need to maintain this package long-term as the webservice logs will reveal how many queries are made using python dataretrieval.

elbeejay:
Aims to close #83 by adding a string to the header in our HTTP GET requests that identifies the python dataretrieval package and version making the request.

Checked functionality against what the R package does and this follows that convention. The "user-agent" information sent to the server is `python-dataretrieval/<version>` which is consistent with the approach in R.

Wrote some tests for this logic.

---

@thodson-usgs can you confirm that my syntax for getting the current version is correct? We don't currently have a `_version.py` file so in the interest of having the code be more resilient there's currently try-except logic to substitute "version-unknown" when it is not available. But I am hoping that once we do the next release the code `dataretrieval.__version__` will successfully return the version information as a string. Used similar logic to automatically set the versions for the hosted documentation that gets built each time a commit is made.

### DIFF
--- a/dataretrieval/utils.py
+++ b/dataretrieval/utils.py
@@ -161,6 +161,13 @@ def set_metadata(response):
     return md
 
 
+def set_useragent():
+    """Function to define a user-agent to include in the GET query."""
+    _version = "CALL FCT TO GET CURRENT VERSION"
+    user_agent = {'user-agent': f"python-dataretrieval/{_version}"}
+    return user_agent
+
+
 def query(url, payload, delimiter=','):
     """Send a query.
 
@@ -188,7 +195,7 @@ def query(url, payload, delimiter=','):
     #    key, value = payload[index]
     #    payload[index] = (key, to_str(value))
 
-    response = requests.get(url, params=payload)
+    response = requests.get(url, params=payload, header=set_useragent())
 
     if response.status_code == 400:
         raise ValueError("Bad Request, check that your parameters are correct. URL: {}".format(response.url))
--- a/dataretrieval/utils.py
+++ b/dataretrieval/utils.py
@@ -2,10 +2,9 @@
 Useful utilities for data munging.
 """
 import warnings
-from importlib.metadata import version
-from importlib.metadata import PackageNotFoundError
 import pandas as pd
 import requests
+import dataretrieval
 from dataretrieval.codes import tz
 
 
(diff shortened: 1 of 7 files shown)


thodson-usgs:
@elbeejay, something is wrong with my understanding. I swore that `setuptools_scm` writes `_version.py` and that infow was read  when calling `dataretrieval.__version__` but now I'm getting an AttributeError. I'll need to debug this further.

elbeejay:
Updated the versioning references based on the documentation for `setuptools_scm`. This resulted in the retirement of support for Python 3.7, as the best practices for both obtaining the package version at runtime, as well as the dynamic loading of version information for the Sphinx documentation (per `setuptools_scm` [documentation](https://pypi.org/project/setuptools-scm/)), require the use of the standard library functionality `importlib.metadata`, which was added in Python 3.8.

dataretrieval/utils.py
```diff
@@ -161,6 +163,16 @@ def set_metadata(response):
     return md
 
 
+def set_useragent():
```
thodson-usgs: move this try-except to `__init__.py` and revise as
```python
try:
    __version__ = version('dataretrieval')
 except PackageNotFoundError:
     __version__ = "version-unknown"
 ```
then `set_useragent` can refer back to `dataretrieval.__version__`

Is that sensible or just extra work?

thodson-usgs (changes requested):
Please revise `__init__.py`

dataretrieval/utils.py
elbeejay: I think that makes sense. I hadn't gone that way because:

1. Modifying `__init__.py` isn't recommended by `setuptools_scm`
2. I couldn't think of any other instance where we'd need to access the version number (since the documentation case is covered in a different way per `setuptools_scm` documentation on best way to do that). 

So I guess with those two items in mind, do you think it's helpful to have the version information accessible at the package level in `__init__.py`? I'm flexible, so happy to make this change if you think it's helpful.

thodson-usgs (approved):
Nice, thanks @elbeejay!

PR #91, opened 2023-03-10
module imports in init

elbeejay:
try to fix relative import issue with init py


thodson-usgs (approved):
Looks great. Thanks, @elbeejay

PR #92, opened 2023-03-15
Stat Service

### LINKED ISSUE #31: Stat Service
Any prediction when the "stat" service will be available? I am getting an error trying to pull annual flow stats from the web indicating that the "stat" service is not yet active

elbeejay:
I didn't see a reason why the request in #31 couldn't be accommodated, so this PR does that.

Might want to wait for us to get Actions back before merging as this fixes the URLs associated with the new repository location and name.

Closes #31


thodson-usgs (approved):
Looks good. I'm noticing some relatively useless helper functions like `_dv` and `_rating`. I don't think you initiated that practice but might be time for me to clean house a bit.

Issue #100, opened 2023-05-06
SSLCertVerificationError


mishranurag:
I tried to download data for a couple of locations and saw the SSLCertVerificationError. I am not sure how to get around this issue.

Sincerely
~A

Edit: Just checked and this happens when I am using a VPN only.

thodson-usgs:
Can't help you with that one. Talk to your IT about getting the correct SSL cert.

mishranurag:
Thank you for the quick response.

SorooshMani-NOAA:
@thodson-usgs recently I started getting SSL certification errors (during automated tests on GitHub for `searvey`) from `dataretireval`. It used to work fine before For examples if I execute:
```python
from dataretrieval import nwis
nwis.get_pmcodes('elevation')
```
I get
```log
/lib/python3.11/site-packages/dataretrieval/nadp.py:44: UserWarning: GDAL not installed. Some functions will not work.
  warnings.warn('GDAL not installed. Some functions will not work.')
Traceback (most recent call last):
  File "/lib/python3.11/site-packages/urllib3/connectionpool.py", line 467, in _make_request
    self._validate_conn(conn)
  File "/lib/python3.11/site-packages/urllib3/connectionpool.py", line 1092, in _validate_conn
    conn.connect()
  File "/lib/python3.11/site-packages/urllib3/connection.py", line 635, in connect
    sock_and_verified = _ssl_wrap_socket_and_match_hostname(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/urllib3/connection.py", line 776, in _ssl_wrap_socket_and_match_hostname
    ssl_sock = ssl_wrap_socket(
               ^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/urllib3/util/ssl_.py", line 466, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/urllib3/util/ssl_.py", line 510, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
[...]

PR #106, opened 2023-07-11
add ssl_check optional parameter

elbeejay:
From #100 - adds an optional boolean parameter `ssl_check` to functions that gets passed through the `requests.get()` to enable skipping of the SSL certificate check. Default value is `True` which checks the SSL certificate (and is the recommended behavior).


thodson-usgs (approved):
great, @elbeejay

PR #107, opened 2023-07-14
Pin numpy<2

thodson-usgs:
ALERT: Numpy 2.0 is coming...


elbeejay:
Ok with me, but I also don't know that we are using any of the functionality that is being retired or broken: https://numpy.org/devdocs/release/2.0.0-notes.html

elbeejay (approved):

Issue #116, opened 2023-10-26
Documentation of function parameters is inconsistent and sometimes misleading for new users


horsburgh:
Using dataretrieval 1.0.6 in my class this week. Students noticed the following in PyCharm:

<img width="1429" alt="image" src="https://github.com/DOI-USGS/dataretrieval-python/assets/3485902/1c917b23-17b6-491d-b32f-2312ac5bfbd2">

I told them they could pass a single site as a string or a list of sites as a list of strings, but they were confused because the highlighted message says "Expected type 'array.pyi' got 'str' instead".  

The documentation of the get_discharge_peaks function says "array of strings":

<img width="619" alt="image" src="https://github.com/DOI-USGS/dataretrieval-python/assets/3485902/dcee2f8f-fe51-454f-bde1-e36ce37c1329">

I even had one student who converted his site code to a Numpy array and passed it to the get_discharge_peaks function. The function actually returns a response, but it's not just for the one site - it had data for lots of sites. Didn't dig in too deep there.

Documentation of the sites parameter for other functions is ambiguous, but not as confusing. In PyCharm:

<img width="863" alt="image" src="https://github.com/DOI-USGS/dataretrieval-python/assets/3485902/29d20928-4feb-4105-bae1-e4e799214dd2">

In the code:

<img width="619" alt="image" src="https://github.com/DOI-USGS/dataretrieval-python/assets/3485902/4ae6f22f-445a-4213-9684-d195aacc9252">

@thodson-usgs - we could take a crack at cleaning this up unless this is intentional or there's another reason not to?

thodson-usgs:
@horsburgh, no that seems odd and the more I look through the code there are several things I'm wondering about.
I have a few suggestions:
1. if possible, start with a smaller PR implementing this for one module. Let me review that and make suggestions before proceeding to others.
2. sites should be a list or list-like (consider adding type checking to enforce that (or using pandas.api.types.is_list_like. This approach feels worse but refer to the doc about which list-like classes should be covered in the type check)
3. It's not totally clear why functions were broken apart like get_iv and _iv, etc? A separate PR could put those back together.
4. Encourage students to start with a unit test, then submit once that test passes.
5. Just in general, smaller more focused PRs make it much easier for me to review things and keep the code base organized.

horsburgh:
[...]

PR #117, opened 2023-11-03
[#116] initial implementation of type hints

pkdash:
This is an initial implementation of adding type hints that includes type hints just for one function of `nwis` module. If this looks good, I can implement the same for the rest of the functions of the `nwis `module.

### DIFF
--- a/dataretrieval/nwis.py
+++ b/dataretrieval/nwis.py
@@ -15,7 +15,7 @@
 import re
 import warnings
 from io import StringIO
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Tuple
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


dataretrieval/nwis.py
```diff
@@ -270,28 +272,31 @@ def _discharge_measurements(ssl_check=True, **kwargs):
     return _read_rdb(response.text), NWIS_Metadata(response, **kwargs)
 
 
-def get_discharge_peaks(sites=None, start=None, end=None,
-                        multi_index=True, ssl_check=True, **kwargs):
+def get_discharge_peaks(sites: Optional[Union[List[str], str]] = None,
+                        start: Optional[str] = None, end: Optional[str] = None,
+                        multi_index: bool = True,
+                        ssl_check: bool = True, **kwargs) -> tuple[pd.DataFrame, BaseMetadata]:
```
thodson-usgs:
```suggestion
                        ssl_check: bool = True, **kwargs) -> Tuple[pd.DataFrame, BaseMetadata]:
```

dataretrieval/nwis.py
```diff
@@ -12,13 +12,15 @@
 
 """
 
+import re
 import warnings
-import pandas as pd
 from io import StringIO
-import re
+from typing import List, Optional, Union
```
thodson-usgs:
```suggestion
from typing import List, Optional, Union, Tuple
```

thodson-usgs (commented):
Excellent @pkdash!
I noticed the 3.8 tests were failing. I suspect `tuple` is only supported for 3.9+.
Please make this change and all tests should pass.

dataretrieval/nwis.py
pkdash: @thodson-usgs Thanks for catching this type hint issue.

PR #122, opened 2023-11-20
Type hints added to public APIs of nwis module

pkdash:
Type hints added to all public functions of the nwis module.


thodson-usgs (approved):
Excellent.
I plan to setup pre-commit to help with formatting. If you haven't used it, I can orient you.
And thank you for your contribution!

PR #128, opened 2023-12-14
[#116] Removing un-necessary helper functions in nwis module

pkdash:


thodson-usgs (commented):
💯  🙏 . Thank you, @pkdash

PR #132, opened 2024-01-26
Add warning message to WQP functions

ehinman:
I created a new function containing the warning message for the upcoming 3.0 profiles in wqp.py. I wasn't quite sure which type of warning we wanted, so I kept the default. I started with the message as a global variable in wqp.py, but then opted to make the warning an internal function. Let me know if another method is preferred (e.g. do we want to read in the txt or...?). 

Tagging @lstanish-usgs so she knows this is in the works!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # dataretrieval: Download hydrologic data
 
-:warning: USGS discrete water quality data availability and format are changing. Beginning in February 2024 the data obtained from legacy profiles will not include new USGS data or recent updates to existing data. To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
-If you have additional questions about these changes, email CompTools@usgs.gov.
+:warning: USGS data availability and format are changing on Water Quality Portal (WQP). Beginning in February 2024 data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data. 
+To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
 
 ## What is dataretrieval?
 `dataretrieval` was created to simplify the process of loading hydrologic data into the Python environment.
--- a/README.md
+++ b/README.md
@@ -16,9 +16,7 @@ though its organization and interface often differ.
 If there's a hydrologic or environmental data portal that you'd like dataretrieval to 
 work with, raise it as an [issue](https://github.com/USGS-python/dataretrieval/issues).
 
-[!WARNING]
-USGS data availability and format are changing. Beginning in February 2024 the data obtained from legacy profiles
-will not include new USGS data or recent updates to existing data. To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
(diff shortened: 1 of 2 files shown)


dataretrieval/wqp.py
```diff
@@ -468,3 +486,19 @@ def _alter_kwargs(kwargs):
     kwargs['mimeType'] = 'csv'
 
     return kwargs
+
+def _three_point_oh_warn():
+    """Private function for warning message about WQX 3.0 profiles
+    """
+
+    warnings.warn('NEWS: USGS data availability and format are changing. '\
```
thodson-usgs: This is pretty verbose. I would just warn that new USGS data will be unavailable for several weeks during the transition to new WQX profiles and refer them to a URL.
No need to add a backslash to the end of the line, so long as your strings are bracketed.
```python
example = (
    'This is a test. '
    'This line will continue the previous one.'
)
```

dataretrieval/wqp.py
```diff
@@ -468,3 +486,19 @@ def _alter_kwargs(kwargs):
     kwargs['mimeType'] = 'csv'
 
     return kwargs
+
+def _three_point_oh_warn():
```
thodson-usgs:
```suggestion
def _warn_v3_profiles_outage():
```
In general, recommend putting "warn" first so like functions are grouped alphabetically.

thodson-usgs (commented):
Thanks @ehinman. Just a couple quick suggestions.

dataretrieval/wqp.py
lstanish-usgs: @ehinman this is the shortened message you can use for the WQP functions: 
NEWS: USGS data availability and format are changing. 
Beginning in February 2024 the data obtained from legacy profiles
will not include new USGS data or recent updates to existing data. 
To view the status of changes in data availability and code functionality, visit:
https://doi-usgs.github.io/dataRetrieval/articles/Status.html
If you have additional questions about these changes, 
email CompTools@usgs.gov.

thodson-usgs: This looks good. Consider adding this warning to the README and note the expected outage in February. A lot of users will be automated scripts or people with older packages, so they likely won't see this in time, but the README might be one of the first places they check.

ehinman:
Thanks for the feedback @thodson-usgs  and @lstanish-usgs. Please let me know if the changes I just made are sufficient. I was reading [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) that adding `[!WARNING]` makes a nice warning box in GitHub, but I'm unsure what it looks like in practice.

thodson-usgs:
From https://gist.github.com/cseeman/8f3bfaec084c5c4259626ddd9e516c61

```
| :warning: WARNING           |
|:----------------------------|
| Another way to warn you     |
```
| :warning: WARNING           |
|:----------------------------|
| Another way to warn you     |

And you can see the effect by clicking "Preview" in the top left of the editor:
image

README.md
```diff
@@ -1,5 +1,8 @@
 # dataretrieval: Download hydrologic data
 
+:warning: USGS data availability and format are changing. Beginning in February 2024 the data obtained from legacy profiles will not include new USGS data or recent updates to existing data. To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
```
ldecicco-USGS: Do we want to specify "USGS discrete water quality data..."? This might get a lot of people unnecessarily stressed

ehinman: That sounds fine to me. Changed in readme and warning message.

Issue #135, opened 2024-03-08
Add NDLI data retrieval functionalities.  


pkdash:
Here are some examples of function to retrieve NDLI data:

- **nldi.get_basin()** - returns a geodataframe with a polygon geometry 
- **nldi.get_flowlines()** - returns a geodataframe with a polyline geometry
- **nldi.get_XYZ()** - returns a geodataframe with a point geometry
- **nldi.search()** - returns a JSON object

thodson-usgs:
Excellent, let's call XYZ `nldi.get_location` or `get_point`. I should've been more careful in my specification, but I meant "XYZ" figuratively as in "to list off several examples".

There are several point datasets within NLDI. I think a single get_point function is fine, but you might choose to split these further into "get_x", get_y", "get_z", etc. :)

Thanks for the update.

thodson-usgs:
The NLDI data retrieval functionality is fully implemented in `dataretrieval/nldi.py` (in main). The module exposes `get_flowlines()`, `get_basin()`, `get_features()`, `get_features_by_data_source()`, and `search()`. An example notebook also exists at `demos/hydroshare/USGS_dataretrieval_NLDI_Examples.ipynb`. Closing as complete.

PR #136, opened 2024-03-08
[#135] NLDI data retrieval

pkdash:
Created this draft PR to get some early feedback. 

This is a draft implementation of 2 functions for accessing NLDI. 
This implementation adds a new dependency - geopandas
NLDI API documentation does not help much in terms of implementing validation for input parameters.
The parameters of the 2 functions that I have implemented probably need more accurate description to make it user friendly.


thodson-usgs:
😮 
```python
gdf = gpd.GeoDataFrame.from_features(feature_collection)
```
...slick. Thank you for showing me that one.

Before the final merge, I want to check the size of the dependencies with GeoPandas, and we might discuss with Jeff whether to make it an optional dependency. A nice feature of `dataretrieval` is that our dependencies are small enough to deploy on Lambda.

Looks great so far.

pkdash:
@thodson-usgs I like the idea of making the GeoPandas as optional dependency. I am going to implement the remaining NLDI data retrieval functions.

thodson-usgs:
@pkdash, did you run out of time? Should I pick this up?

pkdash:
@thodson-usgs Thanks for checking. Yeah has been busy with other projects. I will try to finish it next week. If I can't get to it next week, you can pick this up. I will let you know.

thodson-usgs:
Ah sorry! I'm too used to relying on CI. I should've tested before mucking up your PR. Stand by

thodson-usgs:
@pkdash, I tweaked the CI and build pipeline. I might also tweak how we handle geopandas as an optional dependency.
Otherwise, these changes are looking good so far. Is this PR ready or still draft?

pkdash:
@thodson-usgs The only thing left is handling of the geopandas as optional dependency. I ran out of time yesterday. If you can take care of that then the PR is ready. Thanks.

dataretrieval/nldi.py
```diff
@@ -0,0 +1,495 @@
...
+    # validate the data source
+    _validate_data_source(data_source)
+    url = f'{NLDI_API_BASE_URL}/{data_source}'
+    err_msg = f"Error getting features for data source '{data_source}'"
+    feature_collection = _query_nldi(url, {}, err_msg)
+    gdf = gpd.GeoDataFrame.from_features(feature_collection)
+    return gdf
+
+
+def search(
```
thodson-usgs, line 329: I intend search to be a lower-level function to return a nested dict of multiple feature types. For example, if we wanted to get multiple features upstream of a site, this would allow us to accomplish that in a single API call. Nevertheless, I'm fine with this for the first cut. Great work!

thodson-usgs (approved):
Great work @pkdash! I think I will ultimately rewrite `search`, but otherwise this PR looks good to go.

dataretrieval/nldi.py
pkdash: Due to my lack of understanding of the domain, I was struggling a bit to implement the search functionality.

PR #142, opened 2024-06-27
update(nldi.py): Add CRS information to GeoDataFame objects

jlarsen-usgs:
Set CRS to "EPSG:4326" for GeoDataFrame objects that are returned to the user


thodson-usgs (approved):
Looks good to me. Thank you, @jlarsen-usgs!

PR #143, opened 2024-07-16
 feat(geopandas support): return GeoDataFrame if geopandas is installed

jlarsen-usgs:


dataretrieval/nwis.py
```diff
@@ -71,6 +77,14 @@ def format_response(
     if service == 'peaks':
         df = preformat_peaks_response(df)
 
+    if gpd is not None:
```
thodson-usgs, line 80: Does this remove the original lat/lon column?
If so (and perhaps either way), we should make this conversion optional, so that we don't break existing codes.

thodson-usgs (changes requested):
Very nice. Even better if you added a unit test, but I'll approve it either way.

Otherwise, I had one comment.

dataretrieval/nwis.py
jlarsen-usgs: This maintains the original lat/long columns. This PR just builds a geometry field from that data and creates a geodataframe. I can add a unit test that checks the return type

jlarsen-usgs: Updated tests for geopandas support

thodson-usgs (approved):
looks good!

PR #150, opened 2024-08-07
Remove hardcoded format in nwis.get_discharge_measurements

thodson-usgs:
PR allows users to pass format to `nwis.get_discharge_measurements` while retaining backwards compatibility.

### DIFF
--- a/tests/waterservices_test.py
+++ b/tests/waterservices_test.py
@@ -27,6 +27,7 @@ try:
 except ImportError:
     gpd = None
 
+
 def test_query_waterdata_validation():
     """Tests the validation parameters of the query_waterservices method"""
     with pytest.raises(TypeError) as type_error:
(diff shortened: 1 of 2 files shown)


tests/waterservices_test.py
```diff
@@ -321,10 +323,10 @@ def test_get_discharge_measurements(requests_mock):
 @pytest.mark.parametrize("site_input_type_list", [True, False])
 def test_get_discharge_measurements_sites_value_types(requests_mock, site_input_type_list):
     """Tests get_discharge_measurements method for valid input types for 'sites' parameter"""
-    _format = "rdb"
```
ehinman, line 324: What does the underscore mean? I noticed some tests have _format and others are just format.

thodson-usgs: Typically, it implies that a variable is intended for internal use; i.e. it's used by functions but not directly by the user.
However, in this case, I don't understand why this was done, unless to avoid some naming conflict, but I don't see it.
I'd leave it or now, but sometime it would be good to go back and clean and lint all the tests.

ehinman (approved):
The changes look good to me. I'd really like to pull down your changes and do a quick test on my own machine, but I can't figure out how to do that. Due to the urgency, I'm fine with approving, but if you can tell me how to easily do this for next time, that would be ideal. I tried to clone your fork, but my computer didn't like that I already had the upstream repo in there.

PR #153, opened 2024-08-20
Rename, fix, and extend NAWQA (NWQN) demo

thodson-usgs:
The National Watern Quality Network (NWQN) demo uses AWS serverless to search and pull all NWQN data into an S3 bucket. This PR makes some fixes to the demo and incorporates streamflow.

For context, this is an advanced usage example, which does not currently appear in the doc page. Nevertheless, I host it in the repo for instructing others, but also for helping us to scope development of `dataretrieval` more generally. These pipelines stress several endpoints and help us expose failure modes that appear when we scale up our workflows.


demos/nwqn_data_pull/lithops.yaml
```diff
@@ -8,7 +8,8 @@ aws:
 aws_lambda:
     execution_role: arn:aws:iam::807615458658:role/lambdaLithopsExecutionRole
```
ehinman: Is this execution role specific to your account? Would this be considered sensitive information?

demos/nwqn_data_pull/README.md
```diff
@@ -1,9 +1,11 @@
-# Retrieva data from the National Water Quality Assessment Program (NAWQA)
+# Retrieva data from the National Water Quality Network (NWQN)
 
-This examples walks through using lithops to retrieve data from every NAWQA
+This examples walks through using lithops to retrieve data from every NWQN
```
ehinman:
```suggestion
This example walks through using lithops to retrieve data from every NWQN
```

demos/nwqn_data_pull/README.md
```diff
@@ -1,9 +1,11 @@
...
 monitoring site, then writes the results to a parquet files on s3. Each
```
ehinman:
```suggestion
monitoring site, then writes the results to a parquet file on s3. Each
```

demos/nwqn_data_pull/README.md
```diff
@@ -32,9 +34,11 @@ wget https://www.sciencebase.gov/catalog/file/get/655d2063d34ee4b6e05cc9e6?f=__d
...
-1. Run the script
+1. Run the scripts
 ```bash
-python retrieve_nawqa_with_lithops.py
+python retrieve_nwqn_samples.py
+
+python retrieve_nwqn_streamflow.py
 ```
 
 ## Cleaning up
```
ehinman, line 47: Small typo: lithops

demos/nwqn_data_pull/lithops.yaml
thodson-usgs: No, anyone could use it. It's possible this one requires you to possess USGS AWS credentials, but I can't confirm that without violating policy.

We could replace 80761545858 with <account_id>, but that's not sensitive.

lstanish-usgs: @thodson-usgs can you double check whether this is an internal-only demo? If so then there is likely a better place to put it than in the public-facing package documentation. I suppose we could add a qualifier indicating that USGS credentials are required to use the demo, at a minimum...

thodson-usgs: The instructions indicate that you're supposed to create your own execution_role, but this one will work for USGS employees, at least. I'd prefer to remove the account_id rather than suggest this is internal-only.

(This might not be explicit in the README, but it's part of the setup process in the link)

lstanish-usgs: OK, thanks for clarifying. Removing the account_id sounds like a good approach.

demos/nwqn_data_pull/README.md
```diff
@@ -32,9 +34,11 @@ wget https://www.sciencebase.gov/catalog/file/get/655d2063d34ee4b6e05cc9e6?f=__d
 export DESTINATION_BUCKET=<path/to/bucket>
 ```
 
-1. Run the script
```
ehinman, line 35: I can't seem to comment on unchanged lines, but this refers to line 27: I didn't know I needed to download wget (either in bash or pip install via python) before downloading the sciencebase data using that method. Add a note about it, perhaps.

demos/nwqn_data_pull/retrieve_nwqn_samples.py
```diff
@@ -0,0 +1,174 @@
...
+        def wrapper(*args, **kwargs):
+            attempts = 0
+            while True:
+                try:
+                    return func(*args, **kwargs)
+                except Exception as e:
+                    attempts += 1
+                    if attempts > max_retries:
+                        raise e
+                    wait_time = base_delay * (2 ** attempts)
```
ehinman, line 70: I think I follow to this point: are you making it so that with every failed attempt, the wait time increases exponentially between attempts (until max_retries is satisified)? Might be helpful to add a comment here.

demos/nwqn_data_pull/lithops.yaml
ehinman, line 1: Did not test this functionality. Might make sense to ask someone with an AWS role to try out the lithops piece.

demos/nwqn_data_pull/retrieve_nwqn_samples.py
```diff
@@ -0,0 +1,174 @@
...
+    length_miles = 2 * (drain_area_sq_mi / math.pi) ** 0.5
+    # convert from miles to km
+    return length_miles * 1.60934
+
+
+if __name__ == "__main__":
+    project = "National Water Quality Assessment Program (NAWQA)"
+
+    site_df = pd.read_csv(
+        'NWQN_sites.csv',
```
ehinman, line 163: You could set this directly to the sciencebase URL.

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
```diff
@@ -0,0 +1,134 @@
...
+        site_info = site_info[site_info["agency_cd"] == "USGS"]  # keep only USGS sites
+        site_info = site_info.set_index("site_no")
+
+        main_site = site_info.loc[site]
+        main_site_drainage_area = main_site["drain_area_va"]
+
+        # compute fraction of drainage area
+        site_info = site_info[["drain_area_va"]].copy()
+        site_info["drain_fraction"] = site_info["drain_area_va"] / main_site_drainage_area
+        site_info["fraction_diff"] = np.abs(1 - site_info["drain_fraction"])
```
ehinman, line 64: Do you have any cutoff(s) for drain fraction?

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
```diff
@@ -0,0 +1,134 @@
...
+        # order sites for filling in missing data
+        fill_order = site_info.sort_values("fraction_diff", ascending=True)
+        fill_order = fill_order.index.values
+
+        flow_sites = df.index.get_level_values("site_no").values
+        fill_order = set(fill_order).intersection(flow_sites)
+
+        output = pd.DataFrame()
+
+        # loop through sites, updating the best data first
```
ehinman: "best data" meaning the site with the most similar drainage area?

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
```diff
@@ -0,0 +1,134 @@
...
+        for fill_site in fill_order:
+            fill_data = df.loc[fill_site]
+            output = update_dataframe(output, fill_data)
+
+        output = output.drop(columns=["drain_area_va", "drain_fraction", "fraction_diff"])
+        output["site_no"] = site
```
ehinman, line 86: Should output include metadata indicating when data are from a different basin?

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
```diff
@@ -0,0 +1,134 @@
...
+
```
ehinman, line 87: I tried running this function line-by-line for site '01170095', which had 3 nearby sites: ['01170070', '01170095', '01170100', '01170103']. It appears that the only site with data was '01170070', so that site was wholly used to fill in for site '01170095'. Is this the desired outcome?

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
```diff
@@ -0,0 +1,134 @@
...
+        flow_sites = df.index.get_level_values("site_no").values
+        fill_order = set(fill_order).intersection(flow_sites)
+
+        output = pd.DataFrame()
+
+        # loop through sites, updating the best data first
+        for fill_site in fill_order:
+            fill_data = df.loc[fill_site]
+            output = update_dataframe(output, fill_data)
```
ehinman, line 83: Snazzy!!

ehinman (commented):
Cool stuff! I was able to run the scripts by setting lithops to the LocalhostExecutor, and I used the small "testing" site list. I successfully downloaded several parquet files, though to be honest I haven't gotten to re-opening them and understanding how they're structured. 

It would be nice to overall see more documentation of the code lines and different functions. Though I was able to (mostly) figure out what the code does, this is my first exposure to some of these functions, and I could've gotten to the point faster if there was more narrative on what was going on (for example, I've never seen the `exponential_backoff` method to improve API call handling, and I'd like to know more about the mapping functionality in lithops...is it any more complicated than an "apply" function?).

Do you plan to "fill in" water quality values in some way, similar to streamflow?

Overall, very cool example of using dataretrieval-python with larger data calls.

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
thodson-usgs: yes

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
thodson-usgs: Not necessary for the demo.

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
thodson-usgs: Just that imposed by the neighborhood search.

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
thodson-usgs: Fixing a bug in update_dataframe(). Bots wrote most of this code, so it might take a few iterations to find the bugs.

demos/nwqn_data_pull/README.md
thodson-usgs: Hmm. That will be system-dependent, but I noted that alternatively you can navigate to the url to download the file.

demos/nwqn_data_pull/retrieve_nwqn_streamflow.py
thodson-usgs: yes, revised

thodson-usgs:
> Do you plan to "fill in" water quality values in some way, similar to streamflow?

That process is similar, except we don't rescale by drainage area. For the moment, I think the demo is a successful proof-of-concept for using the NLDI to aggregate data. In practice, we would want several more steps, but I have no plans to implement them at the moment.

Otherwise, this is more or less working as intended and I've addressed the review comments.

demos/nwqn_data_pull/README.md
ehinman: ☝️

ehinman (approved):
Looks good. Thanks!

PR #156, opened 2024-08-30
Adjust code.json to point to new repo location

ehinman:
To make the official release public again, we need to go through the Software Management official release process checklist. I believe we can use most of the info @thodson-usgs provided in [2023](https://code.chs.usgs.gov/software/software-management/-/issues/936), but @ldecicco-USGS noticed that the code.json file still contains the outdated repo info. I have updated it in this PR. Once merged with `master`, I will sync it with GitLab and then make the exact same change to the tagged official release (v1.0.2). Does that sound good?

One question: should we change the metadataLastUpdated (line 42) date in the code.json file?

### DIFF
--- a/code.json
+++ b/code.json
@@ -11,15 +11,15 @@
       "licenses": [
         {
           "name": "Public Domain, CC0-1.0",
-          "URL": "https://code.usgs.gov/cmwsc/shwa/dataretrieval/-/raw/v1.0.2/LICENSE.md"
+          "URL": "https://code.usgs.gov/water/dataretrieval-python/-/raw/v1.0.2/LICENSE.md"
         }
       ]
     },
 
-    "homepageURL": "https://code.usgs.gov/cmwsc/shwa/dataretrieval",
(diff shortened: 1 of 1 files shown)


ldecicco-USGS:
You _might_ be asked to do something like this:
https://code.usgs.gov/water/dataRetrieval/-/blob/main/code.json?ref_type=heads

on the tag - do it as you have done in this PR. Back on master, you can set it up so that it shows the production tag and the development master branch.

thodson-usgs:
@ldecicco-USGS, this is one of the few cases where I might update GitLab first, make sure everything is kosher, then apply the change to GitHub. NBD, either way.

ldecicco-USGS:
Yup, it's specifically a GitLab issue - I probably would too

thodson-usgs (approved):
I approve, though in these cases you may start on GitLab, make sure everything is working and approved there, then push the change back to this PR branch (or main if you're able).

PR #160, opened 2024-09-03
Add info about gwlevels service

ehinman:
Added a couple of sentences about the change to the `gwlevels` service in an attempt to catch/inform users that they need to update their version to leverage the new service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 # dataretrieval: Download hydrologic data
 
-:warning: USGS data availability and format are changing on Water Quality Portal (WQP). Beginning in February 2024 data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data. 
+:warning: USGS data availability and format are changing on Water Quality Portal (WQP). Since March 2024, data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data. 
 To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
+
+:mega: **09/03/2024:** The groundwater levels service has switched endpoints due to WaterServices database updates. `dataretrieval` was updated accordingly in [`v1.0.10`](https://github.com/DOI-USGS/dataretrieval-python/releases/tag/v1.0.10), but older versions of the package will return 503 errors from the discontinued service when using `nwis.get_gwlevels` or the `service='gwlevels'` argument. Visit this Water Data For the Nation [blog post](https://waterdata.usgs.gov/blog/wdfn-waterservices-2024/) for more information.
 
 ## What is dataretrieval?
 `dataretrieval` was created to simplify the process of loading hydrologic data into the Python environment.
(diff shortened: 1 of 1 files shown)


thodson-usgs:
> data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data.
Is the information in the README correct? In separate conversations, I've been told the legacy profiles DO and DO NOT return new USGS data.

thodson-usgs:
Fine to occasionally use the README in this manner, but the principal source for this type of information should be the release notes. I would include your note there: https://github.com/DOI-USGS/dataretrieval-python/releases/v1.0.10

ehinman:
@thodson-usgs the legacy profiles do not include updated/new USGS discrete samples, but they DO include updated/new discrete samples from EPA/STORET.

Also, we recently received an inquiry about the `gwlevels` service returning an error when using `dataretrieval-python`, and it was because they didn't realize that the services had been updated. This is the root of the update to the README. I'm not sure everyone knows to check the release notes, though we could make that big and bold at the top of the readme.

lstanish-usgs:
@ehinman the new warnings look good to me! I agree with @thodson-usgs that the updates to GWL should be in the auto-generated release notes, ideally we could modify those as needed so that they can be interpreted by users in a meaningful way.

thodson-usgs:
> I'm not sure everyone knows to check the release notes, though we could make that big and bold at the top of the readme.

That's fine, and yes, many users don't, but you should always assume this to be the case and be wary of any package that doesn't.

>  ideally we could modify those as needed so that they can be interpreted by users in a meaningful way.

Yes, you can edit them by clicking on the release (link above). I typically start with the auto-generated notes, then rewrite or expand as needed.

ehinman:
Anything else I need to do for this PR? I think it's ready to merge. @thodson-usgs @lstanish-usgs

README.md
```diff
@@ -1,8 +1,10 @@
 # dataretrieval: Download hydrologic data
 
-:warning: USGS data availability and format are changing on Water Quality Portal (WQP). Beginning in February 2024 data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data. 
+:warning: USGS data availability and format are changing on Water Quality Portal (WQP). Since March 2024, data obtained from WQP legacy profiles will not include new USGS data or recent updates to existing data. 
 To view the status of changes in data availability and code functionality, visit: https://doi-usgs.github.io/dataRetrieval/articles/Status.html
 
+:mega: **09/03/2024:** The groundwater levels service has switched endpoints due to WaterServices database updates. `dataretrieval` was updated accordingly in [`v1.0.10`](https://github.com/DOI-USGS/dataretrieval-python/releases/tag/v1.0.10), but older versions of the package will return 503 errors from the discontinued service when using `nwis.get_gwlevels` or the `service='gwlevels'` argument. Visit this Water Data For the Nation [blog post](https://waterdata.usgs.gov/blog/wdfn-waterservices-2024/) for more information.
```
thodson-usgs:
```suggestion
:mega: **09/03/2024:** The groundwater levels service has switched endpoints, and `dataretrieval` was updated accordingly in [`v1.0.10`](https://github.com/DOI-USGS/dataretrieval-python/releases/tag/v1.0.10). Older versions using the discontinued endpoint will return 503 errors for `nwis.get_gwlevels` or the `service='gwlevels'` argument. Visit [Water Data For the Nation](https://waterdata.usgs.gov/blog/wdfn-waterservices-2024/) for more information.
```
Your call here. Mine is a bit tighter, IMO.

thodson-usgs (approved):
looks good. thanks, @ehinman

lstanish-usgs:
@ehinman looks good! Merge away

PR #161, opened 2024-09-06
Update README.md with some badges

thodson-usgs:


jzemmels (approved):
Looks great! 114k downloads and counting is impressive.

thodson-usgs:
> Looks great! 114k downloads and counting is impressive.

Hmm..your approval doesn't seem to allow me to merge. Maybe we need to pull in @ehinman

ehinman (approved):
Hmm we should change this so you don't have to wait for me!

PR #165, opened 2024-09-23
Update code.json to include both dev and official versions

ehinman:
This small MR changes the code.json to point to both the dev and official versions of the repo on GitLab. I would like to change the default repo to "main", but I don't seem to have permissions/maybe this needs to be done manually?

### DIFF
--- a/code.json
+++ b/code.json
@@ -3,8 +3,51 @@
     "name": "dataretrieval",
     "organization": "U.S. Geological Survey",
     "description": "A Python package for discovering and retrieving water data from U.S. federal hydrologic web services.",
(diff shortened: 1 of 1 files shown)


thodson-usgs (approved):
I'll defer to you on this. I assume it is the way it is because that's what the approving official wanted at the time. But that guidance may have been incorrect or changed. Either way, this has no effect on users so there's no risk to changing it.

ehinman:
I switched all references to `master` to `main`. Can you re-review @thodson-usgs?

thodson-usgs (approved):

PR #167, opened 2024-10-04
update nldi service urls

dblodgett-usgs:


thodson-usgs (approved):
looks good, thanks @dblodgett-usgs

PR #168, opened 2024-10-16
Update GH action tags to most recent release commit hashes

jzemmels:
There is a potential security risk where the maintainer of a GitHub Action could retroactively change the code associated with specific, tagged versions of the action (e.g., v1, v2). Laura DeCicco proposed an alternative to replace version numbers in GitHub Action yml files with the specific commit hashes associated with those release versions, which are immutable.

This PR makes those changes for the seven instances where an external action is invoked. I couldn't identify the exact version for some of the actions (e.g., there are multiple v2.#.# versions for the checkout action), so I just selected the hashes associated with the most recently released version for each action.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -19,9 +19,9 @@ jobs:
         python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
 
     steps:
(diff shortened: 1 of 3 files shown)


thodson-usgs (approved):
Yes, I believe that's the recommended practice.

PR #173, opened 2025-03-27
Add samples

ehinman:
This PR adds a function, documentation, testing, etc. for pulling USGS water quality data from the Samples database.

### DIFF
--- a/dataretrieval/samples.py
+++ b/dataretrieval/samples.py
@@ -1,0 +1,213 @@
+"""
+Tool for downloading data from the Water Quality Portal (https://waterqualitydata.us)
+
+See https://waterqualitydata.us/webservices_documentation for API reference
+
+.. todo::
+
+    - implement other services like Organization, Activity, etc.
+
+"""
+
+from __future__ import annotations
+
+import warnings
+from io import StringIO
+from typing import TYPE_CHECKING
+
+import pandas as pd
+
+#from .utils import BaseMetadata, query
+
+if TYPE_CHECKING:
+    from pandas import DataFrame
+
+BASE_URL = "https://api.waterdata.usgs.gov/samples-data/"
+
+services_dict = {
+    "results" : ["fullphyschem", "basicphyschem",
+                    "fullbio", "basicbio", "narrow",
+                    "resultdetectionquantitationlimit",
+                    "labsampleprep", "count"],
+    "locations" : ["site", "count"],
+    "activities" : ["sampact", "actmetric",
+                       "actgroup", "count"],
+    "projects" : ["project", "projectmonitoringlocationweight"],
+    "organizations" : ["organization", "count"]
+}
+
+
+def _check_profiles(
+        service,
+        profile
+):
+    """Check that services are paired correctly with profile in
+    a service call.
+
+    Parameters
+    ----------
+    service : string
+        One of the service names from the "services" list.
+    profile : string
+        One of the profile names from "results_profiles",
+        "locations_profiles", "activities_profiles",
+        "projects_profiles" or "organizations_profiles". 
+    """
+
+    if service not in services_dict.keys():
+        raise TypeError(
+            f"{service} is not a Samples service. "
+            f"Valid options are {list(services_dict.keys())}."
+            )
+    if profile not in services_dict[service]:
+        raise TypeError(
+            f"{profile} is not a profile associated with "
+            f"the {service} service. Valid options are " 
+            f"{services_dict[service]}."
+        )
+
+def get_USGS_samples(
+       service="results",
+       profile="fullphyschem",
+       activityMediaName=None,
+       activityStartDateLower=None,
+       activityStartDateUpper=None,
+       activityTypeCode=None,
+       characteristicGroup=None,
+       characteristc=None,
+       characteristicUserSupplied=None,
+       boundingBox=None,
+       countryFips=None,
+       stateFips=None,
+       countyFips=None,
+       siteTypeCode=None,
+       siteTypeName=None,
+       usgsPCode=None,
+       hydrologicUnit=None,
+       monitoringLocationIdentifier=None,
+       organizationIdentifier=None,
+       pointLocationLatitude=None,
+       pointLocationLongitude=None,
+       pointLocationWithinMiles=None,
+       projectIdentifier=None,
+       recordIdentifierUserSupplied=None
+):
+    """Search Samples database for USGS water quality data.
+    This is a wrapper function for the Samples database API. All potential
+    filters are provided as arguments to the function, but please do not
+    populate all possible filters; leave as many as feasible with their default
+    value (None). This is important because overcomplicated web service queries
+    can bog down the database's ability to return an applicable dataset before
+    it times out.
+
+    The web GUI for the Samples database can be found here:
+    https://waterdata.usgs.gov/download-samples/#dataProfile=site
+
+    If you would like more details on feasible query parameters (complete with
+    examples), please visit the Samples database swagger docs, here:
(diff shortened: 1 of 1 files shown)


ehinman:
@ldecicco-USGS tagging you in case you have some time to take a look and offer feedback on this function. I used the `dataRetrieval` version as inspiration.

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
+"""
+Tool for downloading data from the USGS Aquarius Samples database (https://waterqualitydata.us)
+
+See https://api.waterdata.usgs.gov/samples-data/docs#/ for API reference
+
+Not sure about avilable site types, characteristics, state codes, or other
```
jzemmels:
```suggestion
Not sure about available site types, characteristics, state codes, or other
```

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
...
+        projects - "project", "projectmonitoringlocationweight"
+        organizations - "organization", "count"
+    activityMediaName : string or list of strings, optional
+        Name or code indicating environmental medium in which sample was taken.
+        Example: "Water".
+    activityStartDateLower : string, optional
+        The start date if using a date range. Takes the format YYYY-MM-DD.
+        The logic is inclusive, i.e. it will also return results that
+        match the date. If left as None, will pull all data on or before
+        activityStartDateUpper, of populated.
```
jzemmels: Not sure I understand this sentence -- "of populated" what?

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
...
+        The end date if using a date range. Takes the format YYYY-MM-DD.
+        The logic is inclusive, i.e. it will also return results that
+        match the date. If left as None, will pull all data after
+        activityStartDateLower up to the most recent available results.
+    activityTypeCode : string or list of strings, optional
+        Text code that describes type of field activity performed.
+        Example: "Sample-Routine, regular".
+    characteristicGroup : string or list of strings, optional
+        Characteristic group is a broad category describing one or more
+        of results.
```
jzemmels: Typo?
```suggestion
        results.
```

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
...
+        Example: "Organics, PFAS"
+    characteristc : string or list of strings, optional
```
jzemmels:
```suggestion
    characteristic : string or list of strings, optional
```

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
...
+        with pointLocationLongitude and pointLocationWithinMiles.
+    pointLocationLongitude : float, optional
+        Longitude for a point/radius query (decimal degrees). Must be used
+        with pointLocationLatitude and pointLocationWithinMiles.
+    pointLocationWithinMiles : float, optional
+        Radius for a point/radius query. Must be used with
+        pointLocationLatitude and pointLocationLongitude
+    projectIdentifier : string or list of strings, optional
+        Designator used to uniquely id a data collection project in
+        organization context.
```
jzemmels: What is an "organization context?"  Is there a good example?

dataretrieval/samples.py
```diff
@@ -0,0 +1,283 @@
...
+        results - "fullphyschem", "basicphyschem",
+                    "fullbio", "basicbio", "narrow",
+                    "resultdetectionquantitationlimit",
+                    "labsampleprep", "count"
+        locations - "site", "count"
+        activities - "sampact", "actmetric",
+                        "actgroup", "count"
+        projects - "project", "projectmonitoringlocationweight"
+        organizations - "organization", "count"
+    activityMediaName : string or list of strings, optional
```
jzemmels, line 168: Kinda crazy to me that the only doc pages on this type of parameter are from dataRetrieval or related packages. The [WQP documentation](https://www.waterqualitydata.us/webservices_documentation/#:~:text=the%20data%20elements-,ActivityMediaName,-%2C%20ActivityMediaSubdivisionName%2C%20and%20ResultSampleFractionText) just says something to the effect of "look at the output for more information." Why include it as a query-able parameter if the parameter's possible values aren't communicated to the user? This isn't something to solve in this PR, or even in the package overall, but I feel like this is something seriously lacking in the API documentation we expect users to read. 

I found the "sample media" endpoint through the [codeservice link](https://api.waterdata.usgs.gov/samples-data/codeservice/docs#/sample%20media) you provided that looks to provide the actual values for these more obscure parameters. But pinging one endpoint to learn about the parameter values of a different endpoint seems overly complicated to me. Non-rhetorically: are these parameters so volatile that we _couldn't_ provide static documentation somewhere that describes the values they can take?

jzemmels (approved):
These changes look good to me. The actual function code looks boilerplate (in a good way). I've added some comments/suggestions to the documentation that you might consider addressing/implementing.

dataretrieval/samples.py
ehinman: Ah! "if populated"!

dataretrieval/samples.py
ehinman: I really was on a roll with these typos last week. Thanks for catching.

dataretrieval/samples.py
ehinman: Ah, this was lifted directly from the Samples API descriptions or `dataRetrieval`. I'm going to rewrite and let me see if I can add an example.

dataretrieval/samples.py
ehinman: I think you make great points. I believe @ldecicco-USGS has incorporated some of these types of lookup tables to prevent users from passing invalid options to the API, and there are some basic lookup tables in `dataRetrieval` too. I don't see it as too much trouble to have a set of functions called "[input name]_lookup" that a user can run to return all viable options for Samples. Doubly useful is Laura's approach of checking a user's query before it is made. I will add this as an issue (if it isn't there already) and we can tackle this as a separate PR.

ldecicco-USGS: Yup! There's actually a web service to check available parameters:

https://github.com/DOI-USGS/dataRetrieval/blob/main/R/read_USGS_samples.R#L334

The main bit:
```r
service_options <- c("characteristicgroup", "states", "counties",
                       "countries", "sitetype", "samplemedia",
                       "characteristics", "observedproperty")
  
  check_group_req <- httr2::request("https://api.waterdata.usgs.gov") |> 
    httr2::req_url_path_append("samples-data",
                               "codeservice",
                               service)
```
For instance:
So, https://api.waterdata.usgs.gov/samples-data/codeservice/samplemedia shows all the available sample media options. It's probably not described in the main documentation because (I think) it's dynamically populated from the values in the data base. So, if a data provider suddenly adds "Jello", the full documentation doesn't need a manual update. I think there's a codeservice Swagger somewhere.....

Here it is:
https://api.waterdata.usgs.gov/samples-data/codeservice/docs

ehinman: Thanks Laura, I agree: while some of the options may not change very often, I still think it is a smart move to make the options a service rather than a "hard-coded" list....data managers are adding new PFAS compounds every day, for instance.

This PR links to that codeservice in the samples function documentation, but I do think Joe's point is valid that we can be nicer to the user in python and avoid sending them to the Swagger doc to find all available options. I bet it's a similar number of lines of code in python!

thodson-usgs:
I made some suggestions in the following PR: https://github.com/ehinman/dataretrieval-python/pull/3
In general, please follow PEP8 conventions. While working on this, I also noticed that Jay's unit testing CI doesn't work correctly. We're failing on many tests, yet the action reports them as successful.

PR #183, opened 2025-09-19
Add waterdata infrastructure

ehinman:
This PR will add access to the new water data APIs via the `waterdata` module. 

**CURRENT STATUS 9/19/25**
It is currently a work in progress that appears to work for GET calls in which the user requests one parameter (e.g. site, pcode) at a time. Still working out the POST calls in which a user may request multiple parameters, which requires the use of the CQL2 query language. Stay tuned.

### DIFF
--- a/dataretrieval/waterdata.py
+++ b/dataretrieval/waterdata.py
@@ -93,7 +93,7 @@ def get_daily(
         max_results: Optional[int] = None,
         convertType: bool = True
     ) -> pd.DataFrame:
-    
+
     service = "daily"
     output_id = "daily_id"
 
@@ -106,7 +106,55 @@ def get_daily(
 
     return waterdata_helpers.get_ogc_data(args, output_id, service)
 
-def get_monitoring_locations() -> pd.DataFrame: 
+def get_monitoring_locations(
+        monitoring_location_id: Optional[List[str]] = None,
+        agency_code: Optional[List[str]] = None,
+        agency_name: Optional[List[str]] = None,
+        monitoring_location_number: Optional[List[str]] = None,
+        monitoring_location_name: Optional[List[str]] = None,
+        district_code: Optional[List[str]] = None,
+        country_code: Optional[List[str]] = None,
+        country_name: Optional[List[str]] = None,
+        state_code: Optional[List[str]] = None,
+        state_name: Optional[List[str]] = None,
+        county_code: Optional[List[str]] = None,
+        county_name: Optional[List[str]] = None,
+        minor_civil_division_code: Optional[List[str]] = None,
+        site_type_code: Optional[List[str]] = None,
+        site_type: Optional[List[str]] = None,
+        hydrologic_unit_code: Optional[List[str]] = None,
+        basin_code: Optional[List[str]] = None,
+        altitude: Optional[List[str]] = None,
+        altitude_accuracy: Optional[List[str]] = None,
+        altitude_method_code: Optional[List[str]] = None,
+        altitude_method_name: Optional[List[str]] = None,             
+        vertical_datum: Optional[List[str]] = None,
+        vertical_datum_name: Optional[List[str]] = None,
+        horizontal_positional_accuracy_code: Optional[List[str]] = None,
+        horizontal_positional_accuracy: Optional[List[str]] = None,
+        horizontal_position_method_code: Optional[List[str]] = None,
+        horizontal_position_method_name: Optional[List[str]] = None,
+        original_horizontal_datum: Optional[List[str]] = None,
+        original_horizontal_datum_name: Optional[List[str]] = None,
+        drainage_area: Optional[List[str]] = None,
+        contributing_drainage_area: Optional[List[str]] = None,    
+        time_zone_abbreviation: Optional[List[str]] = None,
+        uses_daylight_savings: Optional[List[str]] = None,
+        construction_date: Optional[List[str]] = None,
+        aquifer_code: Optional[List[str]] = None,
+        national_aquifer_code: Optional[List[str]] = None,
+        aquifer_type_code: Optional[List[str]] = None,
+        well_constructed_depth: Optional[List[str]] = None,
+        hole_constructed_depth: Optional[List[str]] = None,
+        depth_source_code: Optional[List[str]] = None,
+        properties: Optional[List[str]] = None,
+        skipGeometry: Optional[bool] = None,
+        time: Optional[Union[str, List[str]]] = None,
+        bbox: Optional[List[float]] = None,
+        limit: Optional[int] = None,
+        max_results: Optional[int] = None,
+        convertType: bool = True
+        ) -> pd.DataFrame: 
     service = "monitoring-locations"
     output_id = "monitoring_location_id"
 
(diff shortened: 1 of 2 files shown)


dataretrieval/waterdata.py
```diff
@@ -72,6 +74,990 @@
...
+        unit_of_measure: Optional[Union[str, List[str]]] = None,
+        qualifier: Optional[Union[str, List[str]]] = None,
+        value: Optional[Union[str, List[str]]] = None,
+        last_modified: Optional[str] = None,
+        skipGeometry: Optional[bool] = None,
+        time: Optional[Union[str, List[str]]] = None,
+        bbox: Optional[List[float]] = None,
+        limit: Optional[int] = None,
+        max_results: Optional[int] = None,
+        convertType: bool = True
```
agilmore2: how do we pass the api_key API parameter here?

ehinman: I will add documentation about this (still learning the details myself), but your API key should be passed as a header if it exists as an environment variable. This is the line used to grab the api key in one of the helper functions:
`token = os.getenv("API_USGS_PAT")`

So you'll want to get your API key, and then set it using:

```
os.environ["API_USGS_PAT"] = "<your key>"
```
You may need to restart your session to get it to "register".

ehinman: And to be clear: all you need to do is have your key in your environment, you don't need to "set it" in the functions anywhere.

agilmore2:
I see that in _default_headers()
Thanks!

dataretrieval/waterdata_helpers.py
```diff
@@ -1,10 +1,10 @@
 import httpx
 import os
-import warnings
 from typing import List, Dict, Any, Optional, Union
 from datetime import datetime
 import pandas as pd
 import json
+import geopandas as gpd
```
thodson-usgs: We want to keep geopandas as an optional dependency, I think. See the nldi module for example.

dataretrieval/waterdata.py
```diff
@@ -72,6 +74,985 @@
...
+        >>> df = dataretrieval.waterdata.get_daily(
+        ...     monitoring_location_id = ["USGS-05114000", "USGS-09423350"],
+        ...     approval_status = "Approved",
+        ...     time = "2024-01-01/.."
+    """ 
+    service = "daily"
+    output_id = "daily_id"
+
+    # Build argument dictionary, omitting None values
+    args = {
```
jzemmels: Slick

dataretrieval/waterdata.py
```diff
@@ -72,6 +74,985 @@
...
+    # Build argument dictionary, omitting None values
+    args = { 
+        k: v for k, v in locals().items()
+        if k not in {"service", "output_id"} and v is not None
+    }
+
+    return waterdata_helpers.get_ogc_data(args, output_id, service)
+
+def get_monitoring_locations(
+        monitoring_location_id: Optional[List[str]] = None,
```
jzemmels: Do we want to add some argument checking statements, perhaps somewhere in the helpers? Making sure monitoring_location_id is a string, etc. Not sure how difficult this would be to implement, but I'm able to pass seemingly whatever I want to these arguments and a request is still made.

dataretrieval/waterdata.py
```diff
@@ -72,6 +74,985 @@
...
+        aquifer_type_code: Optional[List[str]] = None,
+        well_constructed_depth: Optional[List[str]] = None,
+        hole_constructed_depth: Optional[List[str]] = None,
+        depth_source_code: Optional[List[str]] = None,
+        properties: Optional[List[str]] = None,
+        skipGeometry: Optional[bool] = None,
+        time: Optional[Union[str, List[str]]] = None,
+        bbox: Optional[List[float]] = None,
+        limit: Optional[int] = None,
+        max_results: Optional[int] = None,
```
jzemmels: Similar to a question I brought up in the [dataRetrieval PR](https://github.com/DOI-USGS/dataRetrieval/pull/830#discussion_r2548008797), I'm either not understanding how `limit` and `max_results` are supposed to work or they're not working as intended. The `max_results` argument doesn't seem to impact the number of rows returned in the output. Examples:

`# wd.get_monitoring_locations(site_type_code="GW") # fetches all GW ML ids`
`wd.get_monitoring_locations(site_type_code="GW", max_results = 10) # fetches 10,000 GW ML ids`
`wd.get_monitoring_locations(site_type_code="GW", max_results = 5, limit = 10) # fetches 10 GW ML ids`

tests/waterdata_test.py
jzemmels, line 1: Switching up the monitoring location IDs and parameter codes fetched across these tests would be good. For a future PR: consider making lists of potential argument values and randomly selecting one for each of the test runs.

dataretrieval/waterdata.py
ehinman: This is really good feedback: its behavior is not very logical. Basically the way it's coded is: "If max results is not none/NA, then return just one page of data. If someone specifies the limit, then return the one page with limit number of rows". I've inquired with @ldecicco-USGS about it. Sounds like the "limit" argument causes confusion, so "max_results" was a way to limit the total number of rows returned, as opposed to the total number of rows returned per page. I'm going to do a little more testing on it.

jzemmels: Thanks, that clarifies things a bit. I wonder if `max_results` could be something like `max_pages`. Only setting `limit` will return the full results, `limit` rows per request (I believe this is the intended use of `limit` in the API docs).

If the user sets `max_pages`, then only `max_pages * limit` results will be returned. The user can set `max_pages = 1` to get exactly `limit` rows (as long as `limit <= 10000`). Does that make sense? I think this could be implemented with some minor tweaks to the behavior in `walk_pages`.

dataretrieval/waterdata.py
ehinman: Created an issue: https://github.com/DOI-USGS/dataretrieval-python/issues/188

dataretrieval/waterdata.py
ehinman: Yeah, I bungled the `max_results` argument in the waterdata functions compared to dataRetrieval, and I am going to leave them out for now.

tests/waterdata_test.py
ehinman: I'm going to skip this change for now, as it doesn't test the mechanics of the functions so much as adding some variety/natural noise to what might be returned. The intention is not to check whether the service is up, but to test that the service returns what we'd expect in terms of rows, columns, unique values, etc. Let me know if I misunderstand.

README.md
```diff
@@ -4,123 +4,238 @@
...
-nwis.get_record(sites='03339000', service='dv', start='2017-12-31', parameterCd='00060')
+logging.basicConfig(filename='waterdata.log', level=logging.INFO)
 ```
-...will download daily data with the parameter code 00060 (discharge).
 
-## Accessing the "Internal" NWIS
-If you're connected to the USGS network, dataretrieval call pull from the internal (non-public) NWIS interface.
-Most dataretrieval functions pass kwargs directly to NWIS's REST API, which provides simple access to internal data; simply specify "access='3'".
-For example
+### NWIS Legacy Services (Deprecated but still functional)
```
thodson-usgs, line 122: I'd give a stronger warning here:
```suggestion
### NWIS Legacy Services (Defunct)
```
Deprecation implies everything still works, and warns users about future breaking changes. However, our legacy endpoints are already breaking.

README.md
```diff
@@ -4,123 +4,238 @@
...
-# specify the USGS site code for which we want data.
-site = '03339000'
+```python
+import dataretrieval.waterdata as waterdata
+
+# Get daily streamflow data (returns DataFrame and metadata)
+df, metadata = waterdata.get_daily(
+    monitoring_location_id='USGS-01646500', 
+    parameter_code='00060',  # Discharge
+    time='2024-10-01/2024-10-02'
```
thodson-usgs: The `time` syntax is a little unclear from this example. Does `/` imply a  date range or a list of dates? I think it's worthwile to quickly clarify the functionality here, since most users will have the same question. A simple comment or a second example would suffice.

README.md
mikemahoney218-usgs: Legacy endpoints (with the exception of gwlevels) are still functional and not defunct.

README.md
ehinman: Added another example and clarifying text.

README.md
thodson-usgs: Ah, i was thinking of qwdata, but, right, these are deprecated. What's the latest unset date for daily and iv?

Issue #111, opened 2023-09-20
Add py.typed


davetapley:
I just bumped to 1.0.5 to get https://github.com/DOI-USGS/dataretrieval-python/pull/103.
It works great but VSCode still says:

<img width="622" alt="Screen Shot 2023-09-19 at 6 23 04 PM" src="https://github.com/DOI-USGS/dataretrieval-python/assets/48232/df34fce6-a829-433a-9665-56786d6600c7">

---

I didn't realize you need a `py.typed` file: https://peps.python.org/pep-0561/#packaging-type-information

Good with a PR?

I have a couple of other hints too.

thodson-usgs:
great, @davetapley. Please, go ahead.

thodson-usgs:
Closing as resolved. `py.typed` was added to the package (currently at `dataretrieval/py.typed`) and is included in package data via `pyproject.toml`. Type checkers and IDEs should now recognize the package as typed.

PR #186, opened 2025-11-12
Add py.typed

davetapley:
https://github.com/DOI-USGS/dataretrieval-python/issues/111

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,3 @@
 from setuptools import setup
 
-setup(
-    package_data={"dataretrieval": ["py.typed"]}
-)
+setup()


thodson-usgs (commented):
@davetapley, I think we should configure this in the `pyproject.toml` instead. Something like
```
[tool.setuptools.package-data]
dataretrieval = ["py.typed"]
```
We need to add the py.typed before this PR will work, right?

davetapley:
@thodson-usgs good catch! Updated.

setup.py
```diff
@@ -1,3 +1,5 @@
 from setuptools import setup
 
-setup()
+setup(
+    package_data={"dataretrieval": ["py.typed"]}
+)
```
thodson-usgs:
```suggestion
setup()
```

thodson-usgs: because this is configured in the pyproject.toml now

thodson-usgs (approved):
Thanks for sticking with it!

Issue #185, opened 2025-11-12
PEP 639 license identifier


davetapley:
Per: https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios

For: https://pypi.org/project/licensecheck/ which currently shows:
```
┏━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━
┃ Compatible ┃ Package                   ┃ License
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━
│
│ ✖          │ dataretrieval             │ LICENSE
```

thodson-usgs:
Fixed by #226 (merged). The pyproject.toml now uses `license = "CC0-1.0"` and `license-files = ["LICENSE.md"]` per PEP 639.

PR #187, opened 2025-11-12
PEP 639 license identifier in pyproject.toml

davetapley:
https://github.com/DOI-USGS/dataretrieval-python/issues/185

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,8 @@ description = "Discover and retrieve water data from U.S. federal hydrologic web
 readme = "README.md"
 requires-python = ">=3.8"
 keywords = ["USGS", "water data"]
-license = {file = "LICENSE.md"}
+license = "CC0-1.0"
+license-files = ["LICENSE.md"]
 authors = [
   {name = "Timothy Hodson", email = "thodson@usgs.gov"},
 ]


pyproject.toml
```diff
@@ -8,7 +8,8 @@ description = "Discover and retrieve water data from U.S. federal hydrologic web
 readme = "README.md"
 requires-python = ">=3.8"
 keywords = ["USGS", "water data"]
-license = {file = "LICENSE.md"}
+license = "CC0-1.0"
```
ehinman, line 11: This matches with: https://spdx.org/licenses/CC0-1.0.html

pyproject.toml
```diff
@@ -8,7 +8,8 @@ description = "Discover and retrieve water data from U.S. federal hydrologic web
...
+license-files = ["LICENSE.md"]
```
ehinman, line 12: Not sure this needs the ".md"?

ehinman (commented):
I think this looks fine besides the question of whether ".md" is needed. The other question is why the python 3.8 build tests are failing with the following error:

```
 ValueError: invalid pyproject.toml config: `project.license`.
      configuration error: `project.license` must be valid exactly by one definition (2 matches found):
      
          - keys:
              'file': {type: string}
            required: ['file']
          - keys:
              'text': {type: string}
            required: ['text']
      
      [end of output]
```
I will note that we removed testing of 3.8 in a recent PR. Would you mind pulling in the most recent changes to `main` and testing again?

PR #189, opened 2025-11-25
Fix pathing so new waterdata documentation is available

ehinman:
This small PR fixes the .rst pathing so that the new waterdata documentation appears in the API reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,12 @@ Or using conda:
 conda install -c conda-forge dataretrieval
 ```
 
+To install the "main" branch directly from GitHub, use:
+
+```bash
+pip install git+https://github.com/DOI-USGS/dataretrieval-python.git
+```
+
 ## Usage Examples
 
 ### Water Data API (Recommended - Modern USGS Data)


ehinman:
@jzemmels can you take a quick look at this PR I made that fixes a documentation typo? I want to make sure that info is available before sharing the news of the new waterdata module. Thanks!

README.md
```diff
@@ -49,6 +49,12 @@ Or using conda:
 conda install -c conda-forge dataretrieval
 ```
 
+To install the "main" branch directly from GitHub, use:
+
+```bash
+pip install git+https://github.com/DOI-USGS/dataretrieval-python.git
```
jzemmels, line 55: Maybe not worth a mention, but I don't *think* this works for us since it just uses HTTP. I get a `"SSL certificate problem: self-signed certificate in certificate chain"` that happens any time I try to clone from GitHub with HTTP. And `pip install git+ssh://github.com/DOI-USGS/dataretrieval-python.git` didn't work either.

ehinman: Oh interesting! It works for me. Maybe it's a Mac vs PC thing? Maybe it's a specific environment thing?

PR #190, opened 2025-11-26
Another attempt to fix documentation issues

ehinman:
This PR formats bulleted list and hyperlink docstrings so that they do not error out in the CI/CD pipeline that builds the Sphinx docs.

It also removes testing of python 3.8 and adds testing of python 3.13.


thodson-usgs (approved):
lgtm

PR #193, opened 2025-12-01
Update README.md imports

thodson-usgs:
Changes `import dataretrieval.nwis as nwis` to `from dataretrieval import nwis`, etc.


ehinman (approved):
Looks good to me. Cleaner. Thanks for doing this.

PR #195, opened 2025-12-01
Set waterdata data types

thodson-usgs:
This is an initial PR to imrove data type handling within the `waterdata` module.
Simply calling `pd.to_datetime(df['time'])` appears to correctly handle daily data.

### DIFF
--- a/dataretrieval/waterdata/utils.py
+++ b/dataretrieval/waterdata/utils.py
@@ -680,7 +680,7 @@ def _type_cols(df: pd.DataFrame) -> pd.DataFrame:
     Returns
     -------
     pd.DataFrame
-        The cleaned DataFrame with standardized columns.
+        The DataFrame with columns cast to appropriate types.
 
     """
     cols = set(df.columns)
--- a/dataretrieval/waterdata/utils.py
+++ b/dataretrieval/waterdata/utils.py
@@ -684,7 +684,7 @@ def _type_cols(df: pd.DataFrame) -> pd.DataFrame:
 
     """
     cols = set(df.columns)
-    numerical_cols = ["value", "contributing_drainage_area"]
+    numerical_cols = ["value", "contributing_drainage_area", "drainage_area", "altitude", "altitude_accuracy", "well_constructed_depth", "hole_constructed_depth"]
     time_cols = ["time", "datetime", "last_modified"]
     categorical_cols = [
         "approval_status",
--- a/dataretrieval/waterdata/utils.py
+++ b/dataretrieval/waterdata/utils.py
@@ -668,32 +668,40 @@ def _arrange_cols(
         return df.rename(columns={"id": output_id})
 
 
-def _cleanup_cols(df: pd.DataFrame, service: str = "daily") -> pd.DataFrame:
+def _type_cols(df: pd.DataFrame) -> pd.DataFrame:
     """
-    Cleans and standardizes columns in a pandas DataFrame for water data endpoints.
+    Casts columns into appropriate types.
 
     Parameters
     ----------
     df : pd.DataFrame
         The input DataFrame containing water data.
-    service : str, optional
-        The type of water data service (default is "daily").
 
     Returns
     -------
     pd.DataFrame
         The cleaned DataFrame with standardized columns.
 
-    Notes
-    -----
-    - If the 'time' column exists and service is "daily", it is converted to date objects.
-    - The 'value' and 'contributing_drainage_area' columns are coerced to numeric types.
     """
-    if "time" in df.columns and service == "daily":
-        df["time"] = pd.to_datetime(df["time"]).dt.date
-    for col in ["value", "contributing_drainage_area"]:
-        if col in df.columns:
-            df[col] = pd.to_numeric(df[col], errors="coerce")
+    cols = set(df.columns)
+    numerical_cols = ["value", "contributing_drainage_area"]
+    time_cols = ["time", "datetime", "last_modified"]
+    categorical_cols = [
+        "approval_status",
+        "monitoring_location_id",
+        "parameter_code",
+        "unit_of_measure",
+        ]
+
+    for col in cols.intersection(time_cols):
+        df[col] = pd.to_datetime(df[col], errors="coerce")
+
+    for col in cols.intersection(numerical_cols):
+        df[col] = pd.to_numeric(df[col], errors="coerce")
+
+    for col in cols.intersection(categorical_cols):
+        df[col] = df[col].astype("category")
+
     return df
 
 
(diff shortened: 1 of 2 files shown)


dataretrieval/waterdata/utils.py
```diff
@@ -668,32 +668,40 @@ def _arrange_cols(
...
+        ]
+
+    for col in cols.intersection(time_cols):
+        df[col] = pd.to_datetime(df[col], errors="coerce")
+
+    for col in cols.intersection(numerical_cols):
+        df[col] = pd.to_numeric(df[col], errors="coerce")
+
+    for col in cols.intersection(categorical_cols):
+        df[col] = df[col].astype("category")
```
ehinman: I'm curious whether the "category" type could be pesky for users? What is the benefit?

thodson-usgs: memory efficiency, faster performance, explicit semantics, and occasionaly pesky. I can go either way here. Bottom line, the `waterdata` module is rough, because neither of us really had ocassion to put it into practice. As we do, I expect we'll want to tweak some things.

dataretrieval/waterdata/utils.py
```diff
@@ -668,32 +668,40 @@ def _arrange_cols(
...
-    - The 'value' and 'contributing_drainage_area' columns are coerced to numeric types.
     """
-    if "time" in df.columns and service == "daily":
-        df["time"] = pd.to_datetime(df["time"]).dt.date
-    for col in ["value", "contributing_drainage_area"]:
-        if col in df.columns:
-            df[col] = pd.to_numeric(df[col], errors="coerce")
+    cols = set(df.columns)
+    numerical_cols = ["value", "contributing_drainage_area"]
+    time_cols = ["time", "datetime", "last_modified"]
```
ehinman: Is there a "datetime" column in any of the endpoints? I went through all of them (including continuous) and I don't see it. However, I do see some others:

Time series metadata: `begin`, `end`, `begin_utc`, `end_utc`
Monitoring locations: `construction_date`

And some additional numerics:

Monitoring locations: `altitude`, `altitude_accuracy`, `drainage_area`, `well_constructed_depth`, `hole_constructed_depth`

dataretrieval/waterdata/utils.py
ehinman: In my opinion, formatting numbers and dates make sense, but I'd leave categories for the user to decide.

dataretrieval/waterdata/utils.py
```diff
@@ -668,32 +668,40 @@ def _arrange_cols(
...
-        The cleaned DataFrame with standardized columns.
+        The DataFrame with columns cast to appropriate types.
 
-    Notes
-    -----
-    - If the 'time' column exists and service is "daily", it is converted to date objects.
-    - The 'value' and 'contributing_drainage_area' columns are coerced to numeric types.
     """
-    if "time" in df.columns and service == "daily":
-        df["time"] = pd.to_datetime(df["time"]).dt.date
```
ehinman, line 692: This piece of the function originally was just changing the "time" column to simply a date (not a timestamp) for the **daily values** endpoint only, so that the user wouldn't be confused about whether the value represents a daily aggregated value (min, max, mean, etc.) or a particular measurement. This logic was initially introduced to match what R `dataRetrieval` was doing: https://github.com/DOI-USGS/dataRetrieval/blob/develop/R/walk_pages.R#L141

dataretrieval/waterdata/utils.py
```diff
@@ -668,32 +668,40 @@ def _arrange_cols(
...
-    for col in ["value", "contributing_drainage_area"]:
-        if col in df.columns:
-            df[col] = pd.to_numeric(df[col], errors="coerce")
+    cols = set(df.columns)
+    numerical_cols = ["value", "contributing_drainage_area"]
```
ehinman:
```suggestion
    numerical_cols = ["value", "contributing_drainage_area", "drainage_area", "altitude", "altitude_accuracy", "well_constructed_depth", "hole_constructed_depth"]
```

dataretrieval/waterdata/utils.py
thodson-usgs: Right. It was my recollection that a dt.date lacks datetime functionality, furthermore, the parsing behavior  of pd.to_datetime seems to have changed. By default, it correctly omits the time information. Maybe this was a pandas update, but in the current version, it seems correct to leave "time" as a datetime object.

dataretrieval/waterdata/utils.py
thodson-usgs: I'm not aware of  a `datetime` column, but see no harm in including it.

dataretrieval/waterdata/utils.py
thodson-usgs: Do you have an example where a user might prefer a string to a category?

ehinman: I'm unfamiliar with panda's categories, so I checked out this reference: https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html

It does mention the memory usage savings if there are "few" categories. That makes sense for approval status, but depending upon the user's call, they may have many sites, pcodes, or units of measure. Additionally, it's pretty apparent that breaking up large queries is key for using these endpoints at this time. I wondered how stitching dataframes together affects the categorical type. It is explained here: https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html#merging-concatenation. I could see the categorical getting lost as users patch together their dataframes.

My initial hesitancy comes from the fact that it appears to be similar to R's "factors" (with some key differences), and I have run into issues with factors in the past when I wasn't aware of how they worked. They are useful for statistical packages to recognize how to handle certain columns, but they can act a little bit funky depending upon the operation you're using (and your level of understanding). For example, if you have a numeric category consisting of the factor levels "1", "2", "100", and you try to simply convert them to numeric, you'll get the number corresponding to the internal code of the level, not the actual number the level represents...so they'd probaby be 1, 2, and 3 (see https://cran.r-project.org/doc/FAQ/R-FAQ.html#How-do-I-convert-factors-to-numeric_003f for proper handling). This is different from pandas, where categories can be any dtype, whereas in R they are always strings. Still, it makes me wary of imparting that decision on select columns when we don't know how the data will be used or organized.

dataretrieval/waterdata/utils.py
ehinman: I am not seeing that, or I am misunderstanding. When I run your branch and pull from `get_latest_daily`, the "time" column shows up as "2025-12-01  00:00:00", whereas in the existing implementation, it shows up as "2025-12-01".
```
check, md = waterdata.get_latest_daily(
    monitoring_location_id="USGS-05129115",
    parameter_code="00060"
)
```
I like the existing implementation for daily summaries only, because the date cannot be confused with a singular measurement, and it indeed represents a "summary" value. However, if it causes problems by being inconsisent in the daily summary services, I'm open to applying a consistent rule.

ehinman: Ah, you're correct, `dt.date` does lack datetime functionality. It changes it to an object. Hm. It does make sense to give it a datetime type. Nevermind. We might then just want to say that the additional "00:00:00" added to it doesn't represent a singular value.

thodson-usgs: did it display 00:00:00? it didn't for me, so this behavior was probably changed at some version of pandas.

ehinman (approved):
Now all I can think of is "lgtm". Approved.

Issue #201, opened 2025-12-04
waterdata module - option to remove "superfluous" ID columns and move stable ID columns to end of the dataframe


ehinman:
Every dataframe returned has an "id" column that the `waterdata` module renames to the appropriate [service]_id. Some ids are ephemeral in the database and are meaningless to the user; they cannot be relied upon to pull the same result through time. These include:
"daily_id" from the daily service
"latest_continuous_id" from the latest continuous service
"latest_daily_id" from the latest daily service
"continuous_id" from the continuous service
"field_measurement_id" from the field measurements service

However, some ids are inherently meaningful. These include:
"monitoring_location_id"
"field_visit_id"
"time_series_id"

This issue covers adding an argument arbitrarily set to "True" that removes those superfluous id columns in the first list.

ehinman:
Addressed partially by https://github.com/DOI-USGS/dataretrieval-python/pull/203, but did not remove superfluous columns completely, just moved them to the end.

PR #203, opened 2025-12-10
Add row ordering by `time` and `monitoring_location_id`, if applicable, plus...

### LINKED ISSUE #202: waterdata module - order returned data in a logical way
Currently data from the Water Data APIs comes back with no clear order. R dataretrieval orders columns by time and then monitoring location identifier. We could do the same thing in our module. One thing to keep in mind is that the user may not list "time" or "monitoring_location_identifier" in their properties argument, so we would need to check for that situation before ordering. Here's the internal function written for the R version:

https://github.com/DOI-USGS/dataRetrieval/blob/develop/R/read_waterdata_continuous.R#L101

ehinman:
...move probably-useless-to-user database ids for specific endpoints to the end of the dataframe, so they're less likely to impede review of the data returned.

Also added a couple tests for these changes. Note that the `time` component is sorted in ascending order, so earlier results show up first.

Closes #202 and is a slightly softer solution to #201, since users can request columns in the `properties` argument, anyway.

### DIFF
--- a/dataretrieval/waterdata/utils.py
+++ b/dataretrieval/waterdata/utils.py
@@ -662,9 +662,24 @@ def _arrange_cols(
                 plural = output_id.replace("_id", "s_id")
                 if plural in properties:
                     df = df.rename(columns={"id": plural})
-        return df.loc[:, [col for col in properties if col in df.columns]]
+        df = df.loc[:, [col for col in properties if col in df.columns]]
     else:
-        return df.rename(columns={"id": output_id})
+        df = df.rename(columns={"id": output_id})
+    
+    # Move meaningless-to-user, extra id columns to the end
+    # of the dataframe, if they exist
+    extra_id_cols = set(df.columns).intersection({
+        "latest_continuous_id",
+        "latest_daily_id",
+        "daily_id",
+        "continuous_id",
+        "field_measurement_id"
+        })
+    if extra_id_cols:
+        id_col_order = [col for col in df.columns if col not in extra_id_cols] + list(extra_id_cols)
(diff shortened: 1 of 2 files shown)


dataretrieval/waterdata/utils.py
```diff
@@ -662,9 +662,24 @@ def _arrange_cols(
                 plural = output_id.replace("_id", "s_id")
```
thodson-usgs, line 662: this line confuses me, but the code appears to work as it should

thodson-usgs (approved):
looks good,

dataretrieval/waterdata/utils.py
ehinman: We can revisit, because this might not be necessary. I'll make an issue. Basically, every service returns a straight up "id" column with the data, which is actually different across services. So the package adds the service name to the beginning of the "id" column, e.g. "monitoring_location_id", "daily_id", etc. This part of the function accounts for whether someone enters just "id" into their `properties` argument, or enters "monitoring_locationS_id" (maybe they notice that pattern that it's service + id, and the sites service is called "monitoring-locationS"). If they enter "id", then the resulting dataframe will have the "monitoring_location_id" column name. But if they enter "monitoring_locations_id" (straight up service name, "monitoring locations", plus "id"), then it will return the column name "monitoring_locations_id". I kinda doubt this will be leveraged at all, and adds confusion.

ehinman: https://github.com/DOI-USGS/dataretrieval-python/issues/204

Issue #204, opened 2025-12-11
`_arrange_cols` makes it so that user can specify whether they want a plural ID or a singular ID - needed?


ehinman:
This is confusing behavior, specifically with the `get_monitoring_locations()` function. The service is called "monitoring-locations", but the decision was made to return the "id" column as "monitoring_location_id", NOT "monitoring_locationS_id". However, the behavior is as follows:

- someone could request to have the "id" column returned. In this case, the column returned is named "monitoring_location_id"
- someone could not specify properties at all. In this case, the column returned is named "monitoring_location_id"
- someone could notice the pattern that the "id" columns are renamed to service + "_id" and request to have the "monitoring_locations_id" returned. In this case, the column returned is named "monitoring_locations_id".

Is this a necessary offering? I'm not sure.

Issue #196, opened 2025-12-01
Add metadata collections/lookup tables function to waterdata module


ehinman:
We should make these lookup tables ("Metadata collections") available to users via python, too: https://api.waterdata.usgs.gov/ogcapi/v0/collections?f=html

PR #205, opened 2025-12-19
add reference tables function

ehinman:
This PR adds the reference metadata tables to the package. 

Addresses https://github.com/DOI-USGS/dataretrieval-python/issues/196 and https://github.com/DOI-USGS/dataretrieval-python/issues/204

### DIFF
--- a/.github/workflows/sphinx-docs.yml
+++ b/.github/workflows/sphinx-docs.yml
@@ -18,6 +18,7 @@ jobs:
         shell: bash -l {0}
         run: |
           python -m pip install --upgrade pip
+          pip install "docutils<0.22"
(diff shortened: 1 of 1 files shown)


ehinman:
To do: add tests and deal with GEOPANDAS message.

dataretrieval/waterdata/utils.py
```diff
@@ -547,7 +547,7 @@ def _walk_pages(
     logger.info("Requesting: %s", req.url)
 
     if not geopd:
-        logger.warning(
+        logger.info(
```
thodson-usgs:
```suggestion
        logger.warning(
```
I'd leave this as a warning.

PR #209, opened 2026-01-15
Make ref table function way simpler and more efficient

ehinman:
I realized I initially built the reference table function to match [`dataRetrieval`](https://github.com/DOI-USGS/dataRetrieval/blob/main/R/read_waterdata_metadata.R), thinking that there was a reason we couldn't reuse the `get_ogc_data()` function, but I think this is inefficient and incorrect. Swapped out the repeated code for `get_ogc_data` and changed the "id" column to the singular subject name of the reference table, e.g. "site-type-codes" has the "id" column changed to "site_type_code" (the rest of the columns swap underscores for dashes, so keeping consistent).

I also learned that if there aren't any geometries present in the dataframe, `geopandas` will simply return a `pandas` DataFrame. This avoids the issue I created in the last MR where I thought I had to specify "no geopandas" to get back a regular dataframe, and the logger was yelling about geopandas not being installed due to my oversight.

The only risk is that URLs contain the argument `skipGeometry=FALSE`, like this: 
https://api.waterdata.usgs.gov/ogcapi/v0/collections/time-zone-codes/items?skipGeometry=False&limit=50000

However, it doesn't seem to actually affect what comes back.

### DIFF
--- a/demos/WaterData_demo.ipynb
+++ b/demos/WaterData_demo.ipynb
@@ -87,20 +87,17 @@
    "metadata": {},
    "source": [
     "## Examples\n",
-    "Let's get into some examples using the functions listed above. First, we need to load the `waterdata` module and a few other packages and functions to go through the examples. To run the entirety of this notebook, you will need to install `dataretrieval`, `matplotlib`, and `geopandas` packages. `matplotlib` is needed to create the plots, and `geopandas` is needed to create the interactive maps."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "cd626a14",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "# Install necessary packages to run notebook\n",
+    "Let's get into some examples using the functions listed above. First, we need to load the `waterdata` module and a few other packages and functions to go through the examples. To run the entirety of this notebook, you will need to install `dataretrieval`, `matplotlib`, and `geopandas` packages (plus dependencies). `matplotlib` is needed to create the plots, and `geopandas` is needed to create the interactive maps.\n",
+    "\n",
+    "Note that if you use conda rather than pip, you do not need to install folium and mapclassify separately, as they are included in the conda-forge geopandas install.\n",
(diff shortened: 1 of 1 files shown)


ehinman:
@thodson, @jzemmels, can one of you take a look at this PR this week?

dataretrieval/waterdata/api.py
```diff
@@ -1441,6 +1442,27 @@ def get_reference_table(
...
+    md: :obj:`dataretrieval.utils.Metadata`
+        A custom metadata object including the URL request and query time.
+    
+    Examples
+    --------
+    .. code::
+
+        >>> # Get table of USGS parameter codes
+        >>> ref, md = dataretrieval.waterdata.get_reference_table(
+        ...     collection="parameter-codes
```
jzemmels:
```suggestion
        ...     collection="parameter-codes"
```

pyproject.toml
```diff
@@ -39,13 +39,16 @@ test = [
   "flake8",
 ]
 doc = [
+  "docutils<0.22",
```
jzemmels, line 42: Are these new dependencies from the changes made in this PR? 
I don't see docutils or mapclassify used explicitly in the package files anywhere (from a CTRL  + ALT + F through the package files)

jzemmels (approved):
Changes look good. Just one minor typo in the example and a question about dependencies. Otherwise approved.

pyproject.toml
ehinman: `docutils` is one of the dependencies for building the documentation, and they made some update to the latest version that breaks the pipeline. Keeping it at an earlier version while the other packages catch up seems to fix it for now. 

The other dependencies are needed to use the `gpd.explore()` interactive mapping functionality of `geopandas`, and are not installed if you simply `pip install geopandas`. The waterdata demo notebook I merged in earlier this week uses `gpd.explore()` and currently shows errors in the Jupyter notebook in the documentation: https://doi-usgs.github.io/dataretrieval-python/examples/WaterData_demo.html

PR #230, opened 2026-04-13
Switch WQP default profile from legacy to WQX3.0

### LINKED ISSUE #121: switch default WQP services profile
To be done ~2 months after communications released about the changeover. Update the WQP services functions to default to the basicPhysChem (or narrow?) profile. To ease uses into the change, we might want to add a notification when a user calls the function stating something like: 'retrieving with the default profile (basicPhysChem, or Narrow)'.
Also, just as a reminder, want to make sure this update gets into the changelog.

thodson-usgs:
Closes #121

## Summary
- Changed all WQP function signatures from `legacy=True` to `legacy=False`
- `get_results`, `what_sites`, and `what_activities` now route to WQX3.0 endpoints by default
- Functions without a WQX3.0 equivalent (`what_organizations`, `what_projects`, `what_detection_limits`, `what_habitat_metrics`, `what_project_weights`, `what_activity_metrics`) emit a `UserWarning` when called with the default `legacy=False`, making it clear the user is getting legacy data
- `legacy=True` still works but triggers a `DeprecationWarning` directing users to remove it
- Removed the now-unnecessary `_warn_wqx3_use()` function (WQX3.0 is no longer experimental)

## Behavior change demo (live API)

```python
from dataretrieval.wqp import get_results

# Default: now hits the WQX3.0 endpoint
df, md = get_results(siteid="USGS-01646500", startDateLo="01-01-2024", startDateHi="03-01-2024")
print(md.url)
# https://www.waterqualitydata.us/wqx3/Result/search?siteid=USGS-01646500&...

# Explicit legacy=True still works, with a DeprecationWarning:
df, md = get_results(legacy=True, siteid="USGS-01646500", startDateLo="01-01-2024", startDateHi="03-01-2024")
# DeprecationWarning: ... Setting `legacy=False` will remove this warning.
print(md.url)
# https://www.waterqualitydata.us/data/Result/Search?siteid=USGS-01646500&...
```

## Test plan
- [x] Updated `test_what_sites` and `test_what_activities` to mock WQX3.0 URLs
- [x] Added `test_get_results_legacy` to verify `legacy=True` triggers `DeprecationWarning`
- [x] Live verification: default `get_results` call routes to `/wqx3/Result/search` and returns data; `legacy=True` routes to `/data/Result/Search` and emits the expected `DeprecationWarning`.
- [x] All unit tests pass


thodson-usgs:
Closing this PR after auditing the R `dataRetrieval` package's current stance — we're moving in the opposite direction from upstream.

## R's position (verified against the local clone)

R defaults all WQP functions to **legacy** and treats WQX3 as opt-in beta:

- `R/readWQPdata.R:250` — `service = "Result"` (legacy) is the default; WQX3 routes are `ResultWQX3` / `StationWQX3` / `ActivityWQX3`, opt-in via `service="..."`.
- `R/AAA.R:46-53` — two messaging functions:
  - `wqp_message()` (fires on legacy): *"NEWS: Data does not include USGS data newer than March 11, 2024. More details: https://doi-usgs.github.io/dataRetrieval/articles/Status.html"*
  - `wqp_message_beta()` (fires on WQX3): *"WQX3 services are in-development, use with caution."*
- `vignettes/Status.Rmd:115` makes the policy explicit:
  > *"`dataRetrieval` WQP functions continue to default to the legacy system... The new replacement services aren't currently set as the default because the WQP team still considers these services 'beta', and therefore performance is not guaranteed."*

## What this PR proposed vs. what R does

| | this PR | R |
| --- | --- | --- |
| Default profile | **WQX3** | **legacy** |
| `legacy=True` notice | `DeprecationWarning` | `message()` (legacy is not deprecated) |
| `legacy=False` notice | (removed) | `message()` ("still in development") |
| Frames WQX3 as | "no longer experimental" | "still considers these services 'beta'" |

The "WQX3 is no longer experimental" claim is inconsistent with what the R team — and per them, the WQP team itself — currently believe.

## Recommendation

Hold off on flipping the default until the WQP team marks WQX3 GA. R will likely flip first; we mirror them when it happens. Meanwhile, `main`'s current behavior (default `legacy=True`, both warnings fire) is already roughly aligned.

## Possible small follow-up (separate)

`main`'s `_warn_legacy_use()` uses `DeprecationWarning` and tells users *"Setting `legacy=False` will remove this warning"* — that's stronger than R's plain `message()` and pushes users toward beta services. Worth softening to `UserWarning` and rewording to match R's "data is stale since 2024-03-11" framing. Tracking this as a possible separate PR; not opening one yet pending direction.

PR #232, opened 2026-04-13
Add tsv mimeType support and raise clear error for xlsx in WQP

### LINKED ISSUE #162: Accommodate tsv, xlsx mimeTypes in wqp.py
Only csv is supported by WQX3.0 at this time, but legacy still supports tsv and xlsx. Not a high priority, but something to consider as we build out the api.

thodson-usgs:
Closes #162

## Summary
- `_check_kwargs` now accepts `mimeType='tsv'` alongside `'csv'` for legacy WQP calls
- `mimeType='xlsx'` raises `NotImplementedError` with a clear message (xlsx support is not yet implemented)
- Any other invalid mimeType raises `ValueError: Invalid mimeType. Supported options: 'csv', 'tsv'.`
- Extracted `_read_wqp_response(text, kwargs)` helper that replaces 9 identical `pd.read_csv(..., delimiter=",")` calls — uses `\t` delimiter when `mimeType='tsv'`, `,` otherwise

Note: tsv support applies to legacy WQP calls only. WQX3.0 endpoints only support csv at this time, as noted in the original issue.

## Test plan
- [x] `test_check_kwargs` updated to cover `tsv` (passes), `xlsx` (NotImplementedError), and unchanged cases
- [x] `test_get_results_tsv` added: mocks a tab-delimited response and verifies it is parsed correctly
- [x] All 94 unit tests pass

### DIFF
--- a/dataretrieval/wqp.py
+++ b/dataretrieval/wqp.py
@@ -154,7 +154,7 @@ def get_results(
 
     response = query(url, kwargs, delimiter=";", ssl_check=ssl_check)
 
-    df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
+    df = _read_wqp_response(response.text, kwargs)
     return df, WQP_Metadata(response)
 
 
@@ -208,7 +208,7 @@ def what_sites(
 
     response = query(url, payload=kwargs, delimiter=";", ssl_check=ssl_check)
 
-    df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
+    df = _read_wqp_response(response.text, kwargs)
 
     return df, WQP_Metadata(response)
 
@@ -263,7 +263,7 @@ def what_organizations(
 
     response = query(url, payload=kwargs, delimiter=";", ssl_check=ssl_check)
 
-    df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
+    df = _read_wqp_response(response.text, kwargs)
 
     return df, WQP_Metadata(response)
 
@@ -314,7 +314,7 @@ def what_projects(ssl_check=True, legacy=True, **kwargs):
 
     response = query(url, payload=kwargs, delimiter=";", ssl_check=ssl_check)
 
-    df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
+    df = _read_wqp_response(response.text, kwargs)
 
     return df, WQP_Metadata(response)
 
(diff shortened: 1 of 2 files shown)


thodson-usgs:
Closing this after re-evaluating it against current `main`. This branch is 109 commits behind and every hunk it touches has since been rewritten, so landing it would be a from-scratch reimplementation rather than a merge — worth checking first whether it should exist at all. I don't think it should.

### Measured against the live portal

Legacy WQP does serve all three formats, so the premise of #162 holds:

| mimeType | status | content-type |
| --- | --- | --- |
| `csv` | 200 | `text/csv` |
| `tsv` | 200 | `text/tab-separated-values` |
| `xlsx` | 200 | `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` |

But the getters return a DataFrame, so the wire format is unobservable to callers. csv and tsv parse to identical frames:

```
csv: shape=(2031, 63)   tsv: shape=(2031, 63)
columns identical: True
pd.testing.assert_frame_equal(csv, tsv)  ->  passes
```

Worse, WQX3 **silently ignores** `mimeType=tsv` and returns a comma-delimited body anyway (`Content-Type: text/plain`). This branch has no `legacy` guard, so `_wqp_delimiter()` picks `"\t"` and pandas swallows each whole row as a single field — no exception, no warning:

```
legacy=False + mimeType='tsv'  ->  (2334, 1)
first column: 'Org_Identifier,Org_FormalName,Project_Identifier,Project_Name,...'
```

That is precisely the quiet mangling `_read_wqp_csv` / `_is_code_column` were added to prevent.

### Against current conventions

- `main`'s `_check_kwargs` now says *"Pass mimeType='csv', or omit it -- csv is the only format this package parses."* That was written after this PR opened, under the AGENTS.md error-message policy. This PR reverses it.
- The `NotImplementedError` for `xlsx` conflicts with AGENTS.md's *"Every check raises `ValueError` -- one class for a bad argument value, so a caller catches by shape rather than by which module rejected it."* `main`'s current `xlsx` message is already clearer than the one proposed here.
- `mimeType` is not a term in `CONTEXT.md`, and it is not a *setting* — "a public keyword is not automatically a setting." This would add a legacy-only, mode-dependent public promise for a knob with no observable effect.

`xlsx` fares no better: it would pull in `openpyxl` and still produce the same DataFrame.

### If this comes back

The motivation to look for is malformed WQP CSV quoting breaking real parses — a caller hitting `ParserError` on a free-text field. That would be evidence-backed and would justify a different change (a documented escape hatch, guarded to `legacy=True`). No such report exists today.

Closing #162 alongside for the same reasons.

PR #234, opened 2026-04-14
Add SiteTypeCode and StatisticCode enums for waterdata parameters

### LINKED ISSUE #98: enums for site type, county, etc.
How do we feel about using [`enum`](https://docs.python.org/3.8/library/enum.html) for [site type](https://github.com/DOI-USGS/dataretrieval-python/blob/87351a13b7c7c1cf035e9a359addf289d14f5e50/dataretrieval/nwis.py#L609-L613), [county code](https://github.com/DOI-USGS/dataretrieval-python/blob/87351a13b7c7c1cf035e9a359addf289d14f5e50/dataretrieval/nwis.py#L583-L586C15), etc., instead of `str`?

It would fit nicely with:
- #97

thodson-usgs:
## Summary

- Adds `SiteTypeCode` and `StatisticCode` as `str`-based enums in `dataretrieval/waterdata/enums.py`
- Exports both from `dataretrieval.waterdata` public API
- Both enums extend `str`, so existing callers passing plain strings are unaffected — the enums are purely additive
- `SiteTypeCode` covers all 56 site types from `get_reference_table("site-types")`
- `StatisticCode` covers the 17 named statistic codes (excluding the 90+ numeric percentile entries)

**Usage example:**
```python
from dataretrieval.waterdata import SiteTypeCode, StatisticCode, get_monitoring_locations, get_daily

df, _ = get_monitoring_locations(site_type_code=SiteTypeCode.WELL)
df, _ = get_daily(statistic_id=StatisticCode.MEAN)
```

Closes #98

## Test plan
- [x] `python -m pytest tests/waterdata_test.py` — all 26 tests pass
- [x] Live smoke test: `get_monitoring_locations(state_name="Connecticut", site_type_code=SiteTypeCode.WELL)` returns only `GW` sites
- [x] Verified `requests` serializes enum values correctly (uses native string value, not `__repr__`)

### DIFF
--- a/dataretrieval/waterdata/__init__.py
+++ b/dataretrieval/waterdata/__init__.py
@@ -25,6 +25,7 @@ from .api import (
     get_stats_por,
     get_time_series_metadata,
 )
+from .enums import SiteTypeCode, StatisticCode
 from .types import (
     CODE_SERVICES,
     PROFILE_LOOKUP,
@@ -37,6 +38,8 @@ __all__ = [
     "PROFILES",
     "PROFILE_LOOKUP",
     "SERVICES",
+    "SiteTypeCode",
+    "StatisticCode",
     "get_channel",
     "get_codes",
     "get_continuous",
--- a/dataretrieval/waterdata/__init__.py
+++ b/dataretrieval/waterdata/__init__.py
@@ -32,6 +32,7 @@ from .api import (
     get_stats_por,
     get_time_series_metadata,
 )
+from .enums import SiteTypeCode, StatisticCode
 from .nearest import get_nearest_continuous
 from .ratings import get_ratings
 from .types import (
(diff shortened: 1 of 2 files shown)


thodson-usgs:
Closing as superseded.

Hardcoding `SiteTypeCode` / `StatisticCode` as static enums has been overtaken by the API-driven path on main: `get_codes` (#301) returns the reference tables live, and `get_queryables` (#333) exposes the per-collection parameter domains. A pinned enum list would now be a second source of truth that silently drifts from the service.

The PR's own module docstring points readers at `get_reference_table("site-types")`, which is no longer part of the public API — a good sign of how far the surface has moved since April.

If discoverable constants are still wanted for IDE autocompletion, the better shape is generating or validating them against `get_codes` rather than checking in a hand-maintained table. Branch `fix-enums-site-type` is retained.

PR #241, opened 2026-04-25
Add optional DuckDB connectors prototype (waterdata + wqp)

thodson-usgs:
## Summary

Adds `dataretrieval.duckdb_connectors`, an **optional** extension that wraps DuckDB connections with helper methods exposing the `waterdata` (OGC) and `wqp` endpoints as registerable SQL views. Each helper returns a `duckdb.DuckDBPyRelation`; pass `register_as=<name>` to also publish the result as a named view that subsequent SQL can reference.

This is a prototype / draft — opening for early feedback on shape and coverage before any release. Status: 15 tests passing, `ruff check` + `ruff format --check` clean, demo notebook executes end-to-end against the live API.

## Why

Several use cases get noticeably more ergonomic in SQL than in pandas:

- **Joining heterogeneous endpoints** — site metadata, daily values, time-series metadata, water-quality results all live behind separate getters; once registered as views they JOIN cleanly.
- **Window functions / time aggregations** — top-N flow days per gauge, monthly means, rolling windows.
- **Compose with files on disk** — DuckDB reads Parquet/CSV/S3 natively, so cached water data can be joined with external datasets without leaving SQL.

## What's in the box

```python
from dataretrieval.duckdb_connectors import waterdata, wqp

with waterdata.connect() as wd:
    wd.monitoring_locations(state_name="Illinois", register_as="sites")
    wd.daily(monitoring_location_id=["USGS-05586100"], parameter_code="00060",
             time="2023-01-01/2023-12-31", register_as="daily")
    wd.sql("""
        SELECT s.monitoring_location_name, avg(d.value) AS mean_cfs
        FROM sites s JOIN daily d USING (monitoring_location_id)
        GROUP BY 1 ORDER BY 2 DESC
    """)
```

### Layout

```
dataretrieval/duckdb_connectors/
├── _base.py         # _require_duckdb, _flatten_geometry, _BaseConnection
├── waterdata.py     # WaterdataConnection + connect()
└── wqp.py           # WQPConnection + connect(legacy=, ssl_check=)
```

`dataretrieval/duckdb_connector.py` (singular) is preserved as a backward-compat alias re-exporting the waterdata connector.

### Endpoints exposed

- **waterdata**: `monitoring_locations`, `daily`, `continuous`, `time_series_metadata`, `latest_continuous`, `latest_daily`, `field_measurements`, `samples`
- **wqp**: `get_results`, `what_sites`, `what_organizations`, `what_projects`, `what_activities`, `what_detection_limits`, `what_habitat_metrics`, `what_activity_metrics` (connection holds a `legacy` / `ssl_check` default; per-call overrides supported)
- Generic: `register_table(name, fn, **kwargs)` for any `(DataFrame, metadata)`-returning getter not yet wrapped.

### Optional dependencies

Added to `pyproject.toml`:

```toml
duckdb  = ["duckdb>=1.0.0"]
spatial = ["dataretrieval[nldi]", "dataretrieval[duckdb]"]   # compound extra
```

The DuckDB spatial *extension* is a runtime C++ binary that DuckDB downloads on first `INSTALL spatial` — not a pip package. Pass `spatial=True` to `connect()` to install + load it; registered geometry columns (stored as WKT) can then be parsed with `ST_GeomFromText`.

`pyproject.toml` `tool.ruff.target-version` was bumped to `py310` (waterdata tests already gate at 3.10).

### Geometry handling

Default: GeoDataFrame `geometry` is flattened to WKT plus `longitude`/`latitude` columns so the connectors work without the spatial extension. With `spatial=True`, native `ST_*` functions are available against the same WKT column.

### Demo

`demos/duckdb_waterdata_demo.ipynb` walks through site discovery, daily values, monthly aggregation, top-N window functions, sites×daily join, latest readings, a cross-source waterdata × WQP join, and the spatial-extension affordance.

## Test plan

- [x] `pytest tests/duckdb_connectors_waterdata_test.py tests/duckdb_connectors_wqp_test.py` — 15 passing (mocked at the getter boundary; spatial test skips if extension can't be fetched)
- [x] `ruff check` + `ruff format --check` clean
- [x] Live-tested every notebook section against `api.waterdata.usgs.gov` + `waterqualitydata.us` with an API token
- [x] Backward-compat: `from dataretrieval import duckdb_connector; duckdb_connector.connect()` still works
- [ ] Decide whether `NOTES.md` (design notes captured during prototyping) belongs in the repo or should be dropped before merge
- [ ] Decide whether to widen coverage to `nldi` / `ngwmn`

## Known follow-ups

- Hot-path `import geopandas` cleanup — already applied in this PR; will look for analogous spots elsewhere in the codebase in a follow-up branch.
- DuckDB `ST_Distance_Sphere` axis convention surprised me during the demo; switched the demo to `ST_Within` with a polygon, which is unambiguous. Worth filing upstream once confirmed against duckdb-spatial docs.

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,0 +1,180 @@
+# DuckDB connector — design notes
+
+Notes captured before implementation. The goal is a prototype that lets users
+query USGS waterdata endpoints via DuckDB SQL.
+
+## Package conventions to follow
+
+- **Style**: ruff-managed, py38 target, double quotes, docstring code
+  formatted at width 72.
+- **Type hints**: full hints, use `from __future__ import annotations`,
+  `str | list[str] | None` style.
+- **Docstrings**: numpy-style (Parameters / Returns sections with dashes).
+  Module top has a short one-paragraph summary.
+- **Naming**: snake_case functions, leading underscore for private helpers,
+  UPPER_SNAKE_CASE module constants.
+- **Logging**: `logger = logging.getLogger(__name__)` at module top.
+- **Errors**: raise `ImportError` with a pip install hint when an optional
+  dep is missing, `ValueError` for bad arguments, `RuntimeError` for
+  unexpected response shapes.
+
+## Optional-dependency pattern (mirror `geopandas`)
+
+`dataretrieval/waterdata/utils.py:24`:
+```python
+try:
+    import geopandas as gpd
+    GEOPANDAS = True
+except ImportError:
+    GEOPANDAS = False
+```
+
+For the connector we will:
+- attempt `import duckdb` at the top of the module;
+- on `ImportError` set a sentinel and raise a clear `ImportError` from
+  the public entry point telling users to `pip install dataretrieval[duckdb]`.
+
+`pyproject.toml` extras (current state):
+```toml
+[project.optional-dependencies]
+test = [...]
+doc  = [...]
+nldi = ['geopandas>=0.10']
+```
+Add a new `duckdb = ["duckdb>=1.0.0"]` extra.
+
+## Endpoint shape
+
+All `dataretrieval.waterdata.api.get_*` functions return
+`tuple[pandas.DataFrame | geopandas.GeoDataFrame, BaseMetadata]`. Pagination
+is fully handled inside `_walk_pages`, so a single call is the whole
+result set.
+
+Endpoints we will expose first (highest user value, all OGC):
+- `get_monitoring_locations` — site discovery (returns GeoDataFrame when
+  geopandas installed)
+- `get_daily` — daily values
+- `get_continuous` — instantaneous values (≤3y per call by API contract)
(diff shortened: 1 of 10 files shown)


elbeejay:
I get the overall idea here, but I think there's some value in revisiting the overall purpose and scope of the package. My understanding was that `dataretrieval` is meant to be an atomic Python API wrapper for USGS water data. Every addition of a feature beyond the that base functionality increases the maintenance burden. I'd wonder if something like DuckDB support could be a separate library that wraps the core `dataretrieval`.

thodson-usgs:
nice to hear from you @elbeejay. I agree with you, and I'm leaning against merging this PR. I was curious whether `duckdb` might be useful, but quickly realized that the query language isn't all that important when a bot is writing your queries.

PR #245, opened 2026-05-03
Fix Watershed class: persist instance state and parse JSON

thodson-usgs:
## Summary
- `Watershed.from_streamstats_json` was a classmethod that mutated `cls` instead of building an instance, causing every watershed to share class-level state and the method to return the class itself rather than a `Watershed`.
- `Watershed.__init__(rcode, x, y)` called `get_watershed(...)` and discarded the result, so a freshly-constructed instance had no attributes of its own.
- `get_watershed(format="object")` returned `None` instead of the parsed JSON dict its docstring promised.

This PR makes `from_streamstats_json` return a real instance, makes `__init__` populate `self` via `get_watershed(format="object")`, and makes `format="object"` actually return the parsed JSON. The legacy `_workspaceID` attribute is preserved as a read-only alias for the new `workspace_id`.

While here:
- Dropped the dead `format == "shape"` branch (a stale Fiona-stub `pass` that fell through to a `Watershed` despite documenting a different return type).
- Made the `format` contract explicit: added a `"watershed"` case alongside `"geojson"`/`"object"`, and unrecognized values now raise `ValueError`. Previously any unrecognized format silently fell through to a `Watershed` — the same bug shape that masked the broken `"shape"` branch.
- Fixed `get_sample_watershed` to actually return a `Watershed` instance (preexisting docstring/behavior mismatch — it called `get_watershed(...)` with the default `format="geojson"` and so returned the raw `Response`).
- Switched `json.loads(r.text)` to `r.json()` to match the convention used in `nldi.py`/`nwis.py`.
- Rewrote the `get_watershed` narrative docstring to describe the multi-return contract (`Response` / `dict` / `Watershed` selected by `format`) instead of the previous "always returns a watershed" wording, which contradicted the actual default behavior.

## Minimal reproducible examples

Reconstructed inline because the live StreamStats endpoint is currently returning 404 across requests; the bug shapes are pure-Python and don't depend on the API.

### Bug 1: `from_streamstats_json` mutates the class

```python
class BuggyWatershed:
    @classmethod
    def from_streamstats_json(cls, blob):
        cls.workspace_id = blob["workspaceID"]   # mutates the class
        cls.parameters = blob["parameters"]
        return cls                                # returns the class itself

a = BuggyWatershed.from_streamstats_json({"workspaceID": "WS_A", "parameters": ["a"]})
b = BuggyWatershed.from_streamstats_json({"workspaceID": "WS_B", "parameters": ["b"]})
print(a is BuggyWatershed)                       # True — not an instance!
print(a.workspace_id, b.workspace_id)            # WS_B WS_B — shared class-level state
```

### Bug 2: `__init__` discards the work

```python
class BuggyWatershed:
    def __init__(self, rcode, x, y):
        _ = get_watershed(rcode, x, y)            # result discarded

ws = BuggyWatershed("MA", -72.7, 42.6)
print(hasattr(ws, "workspace_id"))                # False
```

### Bug 3: `format="object"` returns `None`

```python
def buggy_get_watershed(rcode, x, y, format="geojson"):
    if format == "geojson":
        return _response
    elif format == "object":
        pass                                      # implicit None
```

After this PR, `format="object"` returns the parsed `dict`, `Watershed.from_streamstats_json` returns a fresh instance with isolated state, and `Watershed(...)` populates `self`.

## Test plan
- [x] One regression test in `tests/streamstats_test.py` covering the load-bearing behavior change: two watersheds built from different JSON must not share state via class-level attributes (`test_from_streamstats_json_does_not_mutate_class`). The other behaviors — `from_streamstats_json` returns an instance, `format="object"` returns a dict, `__init__` populates `self`, `_workspaceID` resolves — are each one-line obvious fixes; their tests were dropped as trivial.
- [ ] Reviewer: live-API verification was attempted but the StreamStats `watershed.geojson` endpoint is currently returning HTTP 404 for all tested coordinates (service-side issue independent of this PR). Suggested follow-up: re-run the existing `test_get_watershed` integration once StreamStats recovers.

## Follow-up
Filed a separate issue to discuss whether `get_watershed` should always return a `Watershed` (drop the `format` parameter). That would match the function name and the original narrative-docstring promise, but is a breaking change and out of scope for this bug-fix PR.

### DIFF
--- a/dataretrieval/streamstats.py
+++ b/dataretrieval/streamstats.py
@@ -136,26 +136,47 @@ def get_watershed(
         # use Fiona to return a shape object
         pass
 
-    if format == "object":
-        # return a python object
-        pass
-
     data = json.loads(r.text)
+
+    if format == "object":
+        return data
+
     return Watershed.from_streamstats_json(data)
 
 
 class Watershed:
-    """Class to extract information from the streamstats JSON object."""
+    """Class to extract information from the streamstats JSON object.
+
+    Attributes
+    ----------
+    watershed_point : dict
+        GeoJSON feature for the watershed pour point.
+    watershed_polygon : dict
+        GeoJSON feature for the delineated watershed polygon.
+    parameters : list
+        Watershed parameters returned by StreamStats.
+    workspace_id : str
+        StreamStats workspace identifier for the watershed.
+    """
+
+    def __init__(self, rcode, xlocation, ylocation):
+        """Delineate a watershed and populate the instance from the response."""
+        data = get_watershed(rcode, xlocation, ylocation, format="object")
+        self._populate_from_json(data)
 
     @classmethod
     def from_streamstats_json(cls, streamstats_json):
-        """Method that creates a Watershed object from a streamstats JSON."""
-        cls.watershed_point = streamstats_json["featurecollection"][0]["feature"]
-        cls.watershed_polygon = streamstats_json["featurecollection"][1]["feature"]
-        cls.parameters = streamstats_json["parameters"]
-        cls._workspaceID = streamstats_json["workspaceID"]
-        return cls
+        """Construct a :obj:`Watershed` from a StreamStats JSON response."""
+        instance = cls.__new__(cls)
+        instance._populate_from_json(streamstats_json)
+        return instance
 
-    def __init__(self, rcode, xlocation, ylocation):
-        """Init method that calls the :obj:`from_streamstats_json` method."""
-        get_watershed(rcode, xlocation, ylocation)
+    def _populate_from_json(self, streamstats_json):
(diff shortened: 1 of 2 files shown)


thodson-usgs:
Closing as superseded. The `streamstats.Watershed` reliability bugs this PR fixes — the classmethod mutating `cls`/returning the class, and `__init__` discarding the fetched JSON — were repaired on main by #295. On main, `from_streamstats_json` builds a real instance (`cls.__new__` + `_populate`) and `__init__` populates from `json.loads(response.text)`, and the `format=='shape'` branch now raises `NotImplementedError`. The only delta not carried over is the cosmetic `_workspaceID`→`workspace_id` rename, which isn't worth a rebase. Reopen if we want that naming cleanup.

PR #249, opened 2026-05-03
Fix WQP_Metadata.site_info: bind as a real property and forward kwargs

thodson-usgs:
## Summary
`WQP_Metadata.site_info` was defined inside `__init__` as a nested `@property`-decorated function that was never bound to the class. The decorator returned a property descriptor that was immediately discarded when `__init__` returned, so `md.site_info` always fell through to `BaseMetadata.site_info` and raised `NotImplementedError`. The closure also used `parameters[...]` (the captured `**parameters` kwargs) for the lookup but `self._parameters` for the `in` check — an internal inconsistency.

Compounding the issue, every helper called `WQP_Metadata(response)` with no kwargs, so `self._parameters` was always `{}` even after fixing the closure.

This PR:
- Moves `site_info` to a real `@property` on the class.
- Looks up `siteid` first (WQP-native; the previous `sites` / `site` / `site_no` keys reflected an NWIS copy-paste and never matched a WQP query) while keeping the legacy aliases as a fallback.
- Threads `**kwargs` through `WQP_Metadata(response, **kwargs)` from all nine call sites so the property has the parameters it needs.
- Corrects the docstring: the attribute is `comment` (singular, inherited from `BaseMetadata`) and `query_time` is a `datetime.timedelta` (was `datetme.timedelta`).

### Pre-fix bug

```python
import dataretrieval.wqp as wqp

df, md = wqp.what_sites(legacy=True, siteid="USGS-01646500")
md.site_info
# NotImplementedError: BaseMetadata does not implement site_info
```

The pre-fix shape that causes this:

```python
class WQP_Metadata(BaseMetadata):
    def __init__(self, response, **parameters):
        self._response = response
        self._parameters = parameters
        @property                 # decorator runs here
        def site_info(self):      # but the result is a local variable —
            ...                   # it's never attached to the class.
```

So `md.site_info` always resolves to `BaseMetadata.site_info`, which raises.

### Post-fix, live API

```python
import dataretrieval.wqp as wqp

df, md = wqp.what_sites(legacy=True, siteid="USGS-01646500")
print(len(df))                      # 1
result = md.site_info
print(type(result).__name__)        # tuple
df2, md2 = result
print(len(df2))                     # 1
```

## Test plan
- [x] New regression tests in `tests/wqp_test.py`: `site_info` resolves and returns a `(DataFrame, WQP_Metadata)` tuple when `siteid` was provided; returns `None` when no site filter was supplied.
- [x] All `tests/wqp_test.py` (15 tests) pass.
- [x] Live verification against the WQP: `md.site_info` now returns the expected tuple.

## Related PRs

Other open PRs in this bug-review series that touch `dataretrieval/wqp.py` (different functions, no functional conflicts):
- **#250** — preserve user-supplied WQX3.0 `dataProfile` in `get_results`.
- **#256** — polish (warnings instead of print, exception shape, typos, helper extraction).

### DIFF
--- a/dataretrieval/wqp.py
+++ b/dataretrieval/wqp.py
@@ -155,7 +155,7 @@ def get_results(
     response = query(url, kwargs, delimiter=";", ssl_check=ssl_check)
 
     df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
-    return df, WQP_Metadata(response)
+    return df, WQP_Metadata(response, **kwargs)
 
 
 def what_sites(
@@ -210,7 +210,7 @@ def what_sites(
 
     df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
 
-    return df, WQP_Metadata(response)
+    return df, WQP_Metadata(response, **kwargs)
 
 
 def what_organizations(
@@ -265,7 +265,7 @@ def what_organizations(
 
     df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
 
-    return df, WQP_Metadata(response)
+    return df, WQP_Metadata(response, **kwargs)
 
 
 def what_projects(ssl_check=True, legacy=True, **kwargs):
@@ -316,7 +316,7 @@ def what_projects(ssl_check=True, legacy=True, **kwargs):
 
     df = pd.read_csv(StringIO(response.text), delimiter=",", low_memory=False)
 
-    return df, WQP_Metadata(response)
+    return df, WQP_Metadata(response, **kwargs)
 
 
 def what_activities(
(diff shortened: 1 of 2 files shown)


thodson-usgs:
Closing as superseded. The core bug — `site_info` defined as a nested `@property` inside `__init__`, never bound to the class, so `md.site_info` hit `BaseMetadata`'s `NotImplementedError` — was fixed on main by #295 and #305. On main, `WQP_Metadata.site_info` is a real class-level property reading `self._parameters.get('siteid')` and returning `what_sites(...)`, and kwargs are threaded via `WQP_Metadata(response, **kwargs)` in `get_results` and the shared `_what` helper (#320). Only the legacy `sites`/`site`/`site_no` alias fallbacks weren't carried over — minor, not the bug. Reopen if those aliases are still wanted.

PR #250, opened 2026-05-03
Fix get_results: preserve user-supplied WQX3.0 dataProfile

thodson-usgs:
## Summary
The validation block silently overwrote any user-supplied WQX3.0 `dataProfile` because the `else` clause attached to the *outer* `if "dataProfile" in kwargs and ... not in result_profiles_wqx3` and fired whenever that compound condition was False — including the case where the user passed a *valid* profile like `"narrow"` or `"basicPhysChem"`. The documented example `dataProfile="narrow"` therefore did not actually request the narrow profile.

This PR restructures the WQX3.0 branch so the default (`fullPhysChem`) is only applied when the user did *not* set `dataProfile` at all.

It also raises `ValueError` rather than `TypeError` for invalid profile values (these are bad *values*, not bad types) and calls the constructor with a single formatted message — the previous code passed two separate strings, producing a tuple-args exception that printed as `('msg1', 'msg2')`.

## Test plan
- [x] One new regression test in `tests/wqp_test.py` — `test_get_results_wqx3_preserves_user_dataProfile` — confirms a valid user-supplied WQX3.0 profile (`"narrow"`) is preserved end-to-end through the request URL. The accompanying `TypeError` -> `ValueError` shape fix is documented by the diff and not separately unit-tested.
- [x] All `tests/wqp_test.py` (12 tests) pass.
- [x] Full local test suite passes (with `API_USGS_PAT` set).

## Related PRs

Other open PRs in this bug-review series that touch `dataretrieval/wqp.py` (different functions, no functional conflicts):
- **#249** — fix `WQP_Metadata.site_info` property binding.
- **#256** — polish (warnings instead of print, exception shape, typos, helper extraction). Earlier draft of #256 also touched `get_results`'s `TypeError` -> `ValueError` shape; that overlap was reverted so this PR fully owns the `get_results` rewrite.

### DIFF
--- a/dataretrieval/wqp.py
+++ b/dataretrieval/wqp.py
@@ -126,7 +126,7 @@ def get_results(
 
     kwargs = _check_kwargs(kwargs)
 
-    if legacy:
+    if legacy is True:
         valid_profiles = result_profiles_legacy
         kind = "legacy"
         url = wqp_url("Result")
@@ -141,7 +141,7 @@ def get_results(
             f"dataProfile {profile!r} is not a valid {kind} profile. "
             f"Valid options are {valid_profiles}."
         )
-    if not legacy and profile is None:
+    if legacy is not True and profile is None:
         kwargs["dataProfile"] = "fullPhysChem"
 
     response = query(url, kwargs, delimiter=";", ssl_check=ssl_check)
--- a/dataretrieval/wqp.py
+++ b/dataretrieval/wqp.py
@@ -126,29 +126,23 @@ def get_results(
 
     kwargs = _check_kwargs(kwargs)
 
-    if legacy is True:
-        if (
-            "dataProfile" in kwargs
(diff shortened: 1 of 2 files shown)


thodson-usgs:
@copilot-pull-request-reviewer thanks for the review.

**Accepted: `if legacy is True:` for parity** — kept this consistent with the rest of `wqp.py` (e.g., `what_sites` line 207 uses the same idiom). Pushed in `963342d`.

**Declining: regression test for invalid `dataProfile`** — earlier in this review series I trimmed the two `pytest.raises(ValueError, match=...)` tests for invalid profiles per maintainer guidance to omit tests that just re-state a one-line type/message change (the fix is observable from the diff). The `TypeError(msg, msg)` -> `ValueError(single_msg)` change is structural, but the test would just call the function and assert the exception type. The "preserve user-supplied profile" test (which exercises the actual silent-clobber bug) remains.

PR #252, opened 2026-05-03
nldi: fix several user-visible bugs in get_features, get_flowlines, search

thodson-usgs:
## Summary
Six discrete fixes to the NLDI module, bundled because they all touch the same file and individually each is small.

- **`search(find="flowlines")` without `navigation_mode` crashed** with `AttributeError: 'NoneType' object has no attribute 'upper'` from inside `_validate_navigation_mode`. `_validate_navigation_mode` now treats `None` as a clean `ValueError`, returns the upper-cased value (so callers use the normalized form rather than silently sending lowercase), and raises `ValueError` rather than `TypeError` for invalid modes — these are bad values, not bad types. `search` also surfaces a clear error before delegating to `get_flowlines`.
- **`get_flowlines(comid=...)` silently dropped `trim_start`**: the `comid` branch built `query_params` without `trimStart`. The shared params now live outside the branch.
- **`lat=0.0` / `long=0.0` / `comid=0` were treated as missing** because the code used `if lat:` / `if comid:` truthiness. Switched to `is None` checks throughout so the equator/prime meridian / COMID 0 are accepted.
- **The `get_features` error message had an unbalanced quote** (`feature_id 'USGS-X, and data source ...`). Closing quote restored via a small `_features_err_msg` helper that the two branches share.
- **Unconditional `_validate_data_source(feature_source)`** in `get_features` is now guarded by `if feature_source:` so it doesn't validate `None` once #246's cache fix lands.

## Test plan
- [x] Three new regression tests in `tests/nldi_test.py`: `search(find='flowlines')` without nav-mode raises ValueError; `_validate_navigation_mode` raises ValueError on bad input; `_validate_navigation_mode` normalizes lowercase to uppercase. (Tests for the trim_start, falsy-zero, and unbalanced-quote fixes were dropped as trivial.)
- [x] All 14 existing nldi tests still pass.

## Out of scope
The `distance` default mismatch between `get_flowlines` (5km) and `search(distance=50)` is a behavioral debate (which default should win, or should `search` simply pass through when unspecified) and worth a separate, focused PR if desired.

## Related PRs

- **#246** (`nldi-data-source-validation`) fixes a separate bug in `_validate_data_source`'s cache check and adds the **same** `if feature_source:` guard at one line in `get_features`. Trivial textual conflict at merge time; identical fix so either can land first.
- **#258** (`nldi-navigation-mode-valueerror`) is a **strict subset** of this PR (just the `TypeError` -> `ValueError` change in `_validate_navigation_mode`). It can be closed if this PR merges.


thodson-usgs:
@copilot-pull-request-reviewer thanks for the review. Quick note on the one inline comment:

**Declining: `trimstart` vs `trimStart` case sensitivity in the test.**
The assertion `sent.qs.get("trimstart") == ["true"]` actually works as written because `requests_mock`'s `request.qs` attribute is built via `urllib.parse.parse_qs` with default options, which lowercases keys. I verified this locally by running `pytest tests/nldi_test.py::test_get_flowlines_by_comid_includes_trim_start` — it passes. I'll keep the existing assertion (which exercises the parsed-key path) and won't switch to a URL-string assertion.

PR #269, opened 2026-05-06
Add waterdata.get_ratings for USGS stage-discharge ratings via STAC

thodson-usgs:
Adds a Python wrapper around the new USGS Water Data STAC catalog (announced at https://waterdata.usgs.gov/blog/wdfn-rating-curves/) for stage-discharge rating curves. Mirrors R's [`read_waterdata_ratings`](https://github.com/DOI-USGS/dataRetrieval/blob/develop/R/read_waterdata_ratings.R).

## Why a new module

The rating-curve catalog is exposed via STAC (`api.waterdata.usgs.gov/stac/v0/search`) rather than the OGC collections that back every other `waterdata` getter. Two transport-layer differences from OGC:

1. **Filter syntax** — STAC takes a CQL2 string in a `filter` query param. The composer mirrors R's logic: only pin `file_type` server-side when a single value is requested; multi-type requests fetch all matches and filter URLs client-side.
2. **Download path** — STAC features point at a per-rating `.rdb` asset on `stac-files/`. The function downloads each matching asset to a user-supplied `file_path` (default: `tempfile.mkdtemp()`) and parses it with the existing `nwis._read_rdb` helper.

Because of these differences, the implementation lives in `dataretrieval/waterdata/ratings.py` rather than `api.py`.

## Live API examples (verified)

```python
from dataretrieval.waterdata import get_ratings

# Default: exsa ratings for two sites
ratings = get_ratings(
    monitoring_location_id=["USGS-01104475", "USGS-01104460"],
    file_type="exsa",
)
# {'USGS-01104475.exsa.rdb': DataFrame(2001, 4), 'USGS-01104460.exsa.rdb': DataFrame(...)}
ratings["USGS-01104475.exsa.rdb"].head()
#    INDEP  SHIFT  DEP  STOR
# 0   0.00   0.00  0.0   NaN
# 1   0.01  -0.01  0.0   NaN
# 2   0.02  -0.02  0.0   NaN

# Both exsa and corr for the same two sites — 4 tables returned
ratings = get_ratings(
    monitoring_location_id=["USGS-01104475", "USGS-01104460"],
    file_type=["exsa", "corr"],
)

# Bounding-box "what's available?" without downloading
features = get_ratings(
    bbox=[-95.0, 40.0, -92.0, 42.0],
    download_and_parse=False,
)
# 59 STAC feature dicts in this region
```

## Error paths surfaced upfront

```python
get_ratings(monitoring_location_id="USGS-01104475", file_type="bogus")
# ValueError: Invalid file_type ['bogus']. Valid options: ['exsa', 'base', 'corr'].

get_ratings(monitoring_location_id="USGS-01104475", datetime="P7D")
# ValueError: ISO 8601 durations (e.g. 'P7D') are not supported in `datetime`
#             for the rating-curve service. Provide a date or interval instead.
```

The duration check matches a constraint of the upstream rating-curve service that R also enforces client-side.

## Test plan

- [x] **9 unit tests** in `tests/waterdata_ratings_test.py`:
  - Filter composition (single site, multi site, no site, empty).
  - Two error paths (invalid `file_type`, ISO 8601 duration in `datetime`).
  - End-to-end search + download via `requests_mock` (single-site happy path).
  - `download_and_parse=False` returns the raw feature list.
  - Multi-`file_type` correctly omits `file_type` from the server filter and filters URLs locally.
- [x] Live verification against `api.waterdata.usgs.gov/stac/v0/search` for all three demos shown above.

## Related
Part of the v1.1.4 release-staging series; will be linked from a tracking issue.


thodson-usgs:
Tracked in the v1.1.4 release-staging issue: #270.

thodson-usgs:
@ldecicco-USGS, I think this PR is straightforward, but I did refactor the rdb utilities out of the `nwis` module.  Maybe overkill, but low risk.   I'm fine to undo that if you prefer.

ldecicco-USGS (approved):

PR #282, opened 2026-05-17
feat(waterdata): Add resume support for partial chunked queries

thodson-usgs:
**Stacked on top of #280.** This PR's diff includes #280's chunker commit; rebase / squash after #280 merges.

When a chunked OGC `waterdata` call fails partway through (quota exhaustion, mid-pagination 5xx/429, transport error), the chunker now raises `PartialResult` (or its `QuotaExhausted` subclass) carrying the combined partial DataFrame plus a `ChunkManifest` that records how many sub-requests of the cartesian-product plan completed. The same getter accepts the partial metadata back via a new `resume_from=` kwarg; the chunker validates the saved plan matches the fresh args and fetches only the remaining cartesian-product combinations. Callers concatenate their saved partial frame with the resume call's return value to reconstruct the full result.

**Caller flow**

```python
try:
    df, md = get_daily(monitoring_location_id=sites, parameter_code="00060", time="P7D")
except QuotaExhausted as e:
    partial_df, partial_md = e.partial_frame, e.partial_metadata
    # ...wait for hourly window to reset...
    rest_df, rest_md = get_daily(
        monitoring_location_id=sites,
        parameter_code="00060",
        time="P7D",
        resume_from=partial_md,
    )
    df = pd.concat([partial_df, rest_df], ignore_index=True)
```

**Design**

`ChunkManifest` is a frozen dataclass pinning the normalized cartesian-product plan and the completed count. Pinning the plan (not just the input args) lets resume detect when a caller has changed their inputs between the original call and the retry — same-looking args that chunk differently would silently re-fetch wrong sub-ranges. Resume rejects four invalid states with explicit error messages: no manifest on the metadata, current args don't chunk, current plan differs from the saved plan, or the saved manifest is already complete.

`PartialResult` is the new base exception; `QuotaExhausted` now subclasses it and carries the manifest instead of the bare `completed_chunks`/`total_chunks` counts of #280. Any sub-request exception (including PR #279's `_walk_pages` `RuntimeError`) is wrapped via `raise PartialResult(...) from exc`, so the original cause stays accessible via `__cause__`. On a first-chunk-failed scenario, the chunker synthesizes a minimal `requests.Response` carrying just the canonical URL + manifest so caller-side `BaseMetadata.chunk_manifest` always works.

The manifest is also attached to `BaseMetadata.chunk_manifest` on successful chunked calls so callers can log `md.chunk_manifest` to confirm fan-out and observe sub-request count.

**Surface**

- `resume_from: BaseMetadata | None = None` added to all 11 chunked-getter signatures (`get_daily`, `get_continuous`, `get_monitoring_locations`, `get_time_series_metadata`, `get_combined_metadata`, `get_latest_continuous`, `get_latest_daily`, `get_field_measurements`, `get_field_measurements_metadata`, `get_peaks`, `get_channel`).
- New module exports: `ChunkManifest`, `PartialResult`, `_normalize_plan`.

**Tests**

14 new chunker tests cover: manifest properties (total / is_complete / remaining), frozen-dataclass immutability and hashability, normalized-plan order sensitivity, manifest attached on successful chunked calls, no manifest on pass-through, `BaseMetadata.chunk_manifest` round-trip, `PartialResult` wrapping of fetch exceptions with `__cause__` chained, empty-frame handling on first-chunk failure, `partial_metadata` lazy wrapping, resume happy path (skip completed chunks), and four resume rejection paths (mismatched plan / no manifest / already-complete / no chunking). Plus an end-to-end test that exhausts quota partway through, then resumes to complete the query and verifies frame concat reproduces a single-call equivalent.

All 59 chunker / utility tests pass; 42 chunker-specific tests focused.


thodson-usgs:
Superseded by #283, which merged its own resume mechanism (`ChunkInterrupted.call.resume()`). This PR was stacked on the now-closed #280 and would need a full rewrite onto the merged `ChunkedCall` design, and the core resume use case is already covered.

The one capability the manifest approach here adds over the merged in-memory `.call` handle is serializable / cross-process resume. If that's wanted, it's worth filing as a focused follow-up rather than carrying this stack forward.

PR #283, opened 2026-05-18
feat(waterdata): Auto-chunk OGC requests over the URL byte limit

thodson-usgs:
## Summary

The OGC `waterdata` getters (`get_daily`, `get_continuous`, `get_field_measurements`, and the rest of the multi-value-capable functions) previously failed with HTTP 414 once the request URL exceeded the server's ~8 KB byte limit. The common chained-query pattern — pull a long site list from `get_monitoring_locations`, then feed it into `get_daily` — was the main offender.

This PR adds a transparent joint chunker: any over-budget request is fanned out across multiple sub-requests under the hood, and one combined DataFrame is returned. Callers see no API change.

```python
from dataretrieval.waterdata import get_daily, get_monitoring_locations

sites_df, _ = get_monitoring_locations(state_name="Ohio", site_type="Stream")
# Before: HTTP 414 once `sites_df` exceeded ~500 rows.
# After: transparently chunked, one combined DataFrame returned.
df, md = get_daily(
    monitoring_location_id=sites_df["monitoring_location_id"].tolist(),
    parameter_code="00060",
    time="P7D",
)
```

Mirrors R `dataRetrieval`'s [#870](https://github.com/DOI-USGS/dataRetrieval/pull/870), generalized from one filter axis to N joint axes.

## What's new

- **Joint planner**: every multi-value list parameter AND the cql-text `filter` (split on its top-level `OR` clauses) is modeled as a chunkable axis. Greedy halving picks the biggest chunk across all axes until each sub-request URL fits.
- **Resumable interruption**: mid-stream 429 / 5xx / transport failures surface as a typed `ChunkInterrupted` subclass (`QuotaExhausted` or `ServiceInterrupted`) carrying the partial result plus a `.call` handle. Call `exc.call.resume()` once the underlying condition clears — only still-pending sub-requests are re-issued.
- **Pre-emptive quota guard**: after every non-final sub-request, `ChunkedCall` reads `x-ratelimit-remaining` and raises `RequestExceedsQuota` if the rest of the plan won't fit the window. Set `API_USGS_LIMIT=0` to bypass.
- **Shared `requests.Session`**: opened once per chunked call and published via a `ContextVar` so the paginated-loop helpers downstream reuse the same connection pool across every sub-request (saves one TCP/TLS handshake per sub-request after the first).
- **New public types**: `ChunkPlan`, `ChunkedCall`, `RequestTooLarge`, `RequestExceedsQuota`, `ChunkInterrupted` / `QuotaExhausted` / `ServiceInterrupted`, `RateLimited` / `ServiceUnavailable`, `multi_value_chunked` decorator.

## Behavior changes (documented in NEWS.md)

- `BaseMetadata.url` still reflects the user's original query.
- `BaseMetadata.header` now carries the *last* page / sub-request headers (so `x-ratelimit-remaining` is current) rather than the first.
- `BaseMetadata.query_time` is now cumulative wall-clock across every page / sub-request rather than the first page's elapsed.

## Module layout

- New `dataretrieval.waterdata.chunking`: joint planner, exception hierarchy, `ChunkPlan`, `ChunkedCall`, `multi_value_chunked` decorator, shared-session plumbing.
- `dataretrieval.waterdata.utils`: paginated-loop body unified behind a `_paginate(req, parse_response, follow_up, client)` strategy helper shared by `_walk_pages` and `get_stats_data`.
- `dataretrieval.waterdata.filters`: slimmed to a CQL-parsing leaf; URL-budget and filter-chunking logic moved into the joint planner.

## Test plan

- [x] 51 new unit tests in `tests/waterdata_chunking_test.py` covering the planner, axis extraction, cartesian-product enumeration, rate-limit gating, resume idempotency and equivalence, transient-error classification, shared-session reuse.
- [x] URL-construction stress test against the real `_construct_api_requests` builder: 500 USGS site IDs × 20 datetime OR-clauses, asserting every sub-request URL stays under 8000 bytes.
- [x] Regression tests for the pre-merge code review's bug fixes (empty-frame GDF preservation, single-frame aliasing, `iter_sub_args` passthrough copy, quota check on resume, `RequestExceedsQuota.call` handle, missing-features defensiveness).
- [x] Mid-pagination 429 / 5xx covered for both the OGC and stats paginators.
- [x] CI matrix (Ubuntu / Windows × Python 3.9 / 3.13 / 3.14) all green.
- [x] End-to-end live API verification against `get_monitoring_locations(state_name="Ohio") → get_daily(...)`: 2,888 sites → ~52 KB of comma-joined IDs (6.5× the byte limit) → transparently chunked, 1,444 rows returned in ~7s.


dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,568 @@
...
+~8 KB URL byte limit. This module adds a single decorator that plans
+both chunking dimensions together and iterates the joint cartesian
+product so each sub-request URL fits.
+
+Design:
+
+- Chunkable dims include multi-value list params (sites, parameter
+  codes, ...) and the cql-text ``filter`` (split at top-level ``OR``
+  to keep each chunk valid CQL).
+- The planner enumerates candidate filter chunk counts
```
thodson-usgs: what is k here?

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,568 @@
...
+    sub-request), or the joint cartesian-product plan would issue
+    more than ``max_chunks`` sub-requests."""
+
+
+class QuotaExhausted(RuntimeError):
+    """Raised mid-chunked-call when the API's reported remaining quota
+    (``x-ratelimit-remaining`` header) drops below the configured safety
+    floor. The chunker stops before issuing the next sub-request to
+    avoid a mid-call HTTP 429 that would silently truncate paginated
+    results.
```
thodson-usgs: This seems like a bug. A mid-call HTTP 429 should not silently truncate. If it does, fix it, then we won't need to defend against this case.

dataretrieval/waterdata/filters.py
```diff
@@ -1,47 +1,38 @@
...
 import requests
 
 FILTER_LANG = Literal["cql-text", "cql-json"]
 
-# Conservative fallback budget when ``_chunk_cql_or`` is called without
-# an explicit ``max_len``. The ``chunked`` decorator computes a tighter
-# per-request budget from ``_WATERDATA_URL_BYTE_LIMIT``.
-_CQL_FILTER_CHUNK_LEN = 5000
-
 # Empirically the API replies HTTP 414 above ~8200 bytes of full URL —
```
thodson-usgs: Should this be moved to a different module now?

thodson-usgs (commented):
Pay close attention to the layout: are all variables and functions placed logically into their modules? Or has the logic been mixed up.

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
+"""Joint URL-byte chunking for the Water Data OGC getters.
+
+PR 233 routes most services through GET with comma-separated values
+(e.g. ``monitoring_location_id=USGS-A,USGS-B,...``). Long lists and
+long top-level-``OR`` CQL filters can independently blow the server's
+~8 KB URL byte limit. This module adds a single decorator that plans
+both chunking dimensions together and iterates the joint cartesian
+product so each sub-request URL fits.
```
thodson-usgs: Tighten this to summarize the correct design. Don't mention PR 233 or the old design.

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
...
+        "skip_geometry",
+    }
+)
+
+# Default cap on the number of sub-requests a single chunked call may
+# emit. The USGS Water Data API rate-limits each HTTP request
+# (including pagination), so the true budget is
+# ``hourly_quota / avg_pages_per_chunk``. 1000 matches the default
+# hourly quota.
+_DEFAULT_MAX_CHUNKS = 1000
```
thodson-usgs: why not just set this as x-ratelimit-remaining by default

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
...
+
+# When ``x-ratelimit-remaining`` drops below this between sub-requests,
+# the chunker bails with ``QuotaExhausted`` rather than risk a mid-call
+# HTTP 429. Carries the partial result so callers can resume from a
+# known offset instead of retrying the whole chunked call from scratch.
+_DEFAULT_QUOTA_SAFETY_FLOOR = 50
```
thodson-usgs: after the retrieving the first chunk, just check x-ratelimit-remaining, and if the plan will not fit within our current rait limit, bail and return an Error message that the query would exceed our rate limit and report by how much.

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
...
+    """Raised mid-chunked-call when the API's reported remaining quota
+    (``x-ratelimit-remaining`` header) drops below the configured safety
+    floor. The chunker stops before issuing the next sub-request and
+    surfaces the partial result so callers can resume after the hourly
+    window resets.
+
+    A bare 429 raised by ``_walk_pages`` would also abort the call but
+    discard the chunks completed so far; this exception is the
+    structured-recovery alternative, triggered pre-emptively while the
+    accumulated frames are still in hand.
```
thodson-usgs: Revise this. This error should raise if we prematurely exhaust our quota because another processes is using it faster than predicted. A single process should never incounter this, because it raises a RequestExceedsQuota value error after checking the limit in the first chunk

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
...
+    *,
+    build_request: Callable[..., Any],
+    url_limit: int | None = None,
+    max_chunks: int | None = None,
+    quota_safety_floor: int | None = None,
+) -> Callable[[_FetchOnce], _FetchOnce]:
+    """Decorator that splits multi-value list params and cql-text
+    filters across sub-requests so each sub-request URL fits
+    ``url_limit`` bytes (defaults to ``_WATERDATA_URL_BYTE_LIMIT``)
+    and the joint cartesian-product plan stays ≤ ``max_chunks``
```
thodson-usgs: again max_chunks should be replaced by the our current rate limit (which we know after receiving the first chunk)

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,623 @@
...
+    affect this decorator uniformly.
+
+    ``url_limit`` is enforced against total request bytes (URL + body,
+    via ``_request_bytes``). POST routes are conservatively sized.
+
+    Between sub-requests the wrapper reads ``x-ratelimit-remaining`` from
+    each response. If it drops below ``quota_safety_floor`` (default
+    ``_DEFAULT_QUOTA_SAFETY_FLOOR``), the wrapper raises
+    ``QuotaExhausted`` carrying the combined partial result and the
+    chunk offset so callers can resume after the hourly window resets.
```
thodson-usgs: This should also change. There is no quota_saftey_floor anymore. But raise the QuotaExhausted if we receive an HTTP error telling us the quota was exhuasted.

thodson-usgs (commented):
Please fix my comments

thodson-usgs:
Pushed `01e579e` reworking the quota machinery per all the inline comments. Summary:

**Module docstring (line 8 comment):** tightened. No more PR 233 / two-decorator references — it just describes the current design (joint planner + dynamic quota gate).

**`_DEFAULT_MAX_CHUNKS` / `_DEFAULT_QUOTA_SAFETY_FLOOR` (line 95, 101 comments):** both deleted. After the first sub-request the wrapper reads `x-ratelimit-remaining` directly; if the remaining plan won't fit in the current window, it raises a new `RequestExceedsQuota(ValueError)` carrying `planned_chunks`, `available`, and `deficit` so the message reports exactly how far over budget the call is.

**`QuotaExhausted` (line 144 comment):** rewritten. It now fires only when an actual HTTP 429 propagates from a sub-request (detected by walking `__cause__` for `RuntimeError("429: ...")`, the shape `_raise_for_non_200` produces). The docstring states explicitly that a single-process caller should not normally see this — `RequestExceedsQuota` short-circuits in chunk 1; arrival here implies a concurrent consumer drained the bucket faster than predicted. Carries the partial frame for resume. `partial_response` is `None` when the 429 hits chunk 0.

**`max_chunks` / `quota_safety_floor` decorator params (line 552, 564 comments):** removed. `_plan_list_chunks` and `_plan_joint` no longer carry a `max_chunks` cap; `RequestTooLarge` fires only on the genuine "nothing left to split" floor. The rate-limit gate replaces the static cap.

**Side effects:** `_fetch_once` in `utils.py` already called the decorator with defaults only, so no call-site changes were needed. Ten obsolete cap/floor tests were removed and eight new ones added covering `RequestExceedsQuota` after chunk 0, deficit reporting, the no-header skip path, mid-call 429 → `QuotaExhausted` with partial frame, the first-chunk 429 (`partial_response=None`) edge case, and non-429 `RuntimeError` pass-through. All chunker tests + offline stress test pass.

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,692 @@
...
+already fit get a trivial single-step plan — the wrapper has one code
+path either way.
+
+Planning: for a filter with ``n_clauses`` top-level OR clauses, try
+candidate filter chunk counts ``k = 1, 2, 4, ..., n_clauses``. For
+each, partition clauses into ``k`` count-balanced groups joined by
+``OR``, take the longest (URL-encoded) group as the worst-case filter,
+then plan list-dim chunking by greedy halving against the remaining
+budget. Keep the candidate with the smallest ``list_count × k``.
+
```
thodson-usgs: this sounds like filter chunking is given priority over list-dim chunking. Shouldn't the filter be treated as any other dim?

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,692 @@
...
+                else:
+                    yield {**base, _FILTER_KEY: filter_chunk}
+
+    def execute(self, fetch_once: _FetchOnce) -> tuple[pd.DataFrame, requests.Response]:
+        """Run the plan and return the combined result. See
+        ``_ChunkExecution`` for the per-sub-request semantics."""
+        return _ChunkExecution(self, fetch_once).run()
+
+    @classmethod
+    def from_args(
```
thodson-usgs: I think it would be cleaner to refactor this to a class.init

dataretrieval/waterdata/chunking.py
```diff
@@ -0,0 +1,692 @@
...
+    head = responses[0]
+    if len(responses) > 1:
+        head.headers = responses[-1].headers
+        head.elapsed = sum((r.elapsed for r in responses[1:]), start=head.elapsed)
+    if canonical_url is not None:
+        head.url = canonical_url
+    return head
+
+
+class _ChunkExecution:
```
thodson-usgs: Could we call this QueryExecutor? Really a class should be a noun. So QueryOrchestrator or ChunkManager or ChunkExecutor

thodson-usgs (commented):
fix these things

PR #285, opened 2026-05-21
feat(waterdata): Add async parallel chunker over httpx.AsyncClient

thodson-usgs:
## Summary

Adds an **async-only** parallel chunker to the Water Data getters. When a multi-value query (many monitoring locations, many parameter codes, large CQL2 filter) would exceed the server's ~8 KB URL/body limit, the call is split into multiple sub-requests and run concurrently under one shared `httpx.AsyncClient` with the connection pool sized from `API_USGS_CONCURRENT` (default **16**).

Mid-stream transients (429 / 5xx / connect/read timeout / oversize follow-up URL) escalate to a resumable `ChunkInterrupted` carrying `.call`; the user calls `exc.call.resume()` once the underlying condition clears and only the still-pending sub-requests are re-issued.

Rebased onto current `main` after the merge of #289 (httpx), #283 (auto-chunking), #290 (notebook migration), #291 (vignette ports), #292 (compact CQL2 JSON), and **#293 (ruff-only pre-commit + AGENTS.md + housekeeping)**.

## Commit structure (2 commits)

| | |
|---|---|
| `feat(waterdata): async parallel chunker over httpx.AsyncClient` | All chunker / retry / resume / progress / finalize / deps work, squashed. |
| `chore(pre-commit): add nbstripout to strip notebook outputs on commit` | Stacks on #293's ruff-only config; adds just the `kynan/nbstripout` hook (the one piece #293 didn't cover) and strips outputs from the 15 checked-in waterdata notebooks once. |

## Benchmark

**~6.7× speedup** on a 6-state `get_daily` discharge pull over 17,298 stream sites (chunked into 64 sub-requests): parallel @ `API_USGS_CONCURRENT=16` → **3.11 s**; serial @ `=1` → **20.94 s** (~1,920 rows each, distinct date windows per side so the server cache can't bias either run; both authenticated via `API_USGS_PAT`).

## `API_USGS_CONCURRENT`

| Value | Behavior |
|---|---|
| unset / blank | parallel, cap = **16** (default) |
| `≥ 2` | parallel, connection-pool capped at that value |
| `1` | single-connection gather (one request at a time) |
| `unbounded` | no per-call cap — caller owns the burst risk |
| `0`, negative, malformed | `ValueError` at call time |

Concurrency is bounded **by the connection pool** (`httpx.Limits(max_connections=N, max_keepalive_connections=N)`), not a semaphore. The same pool is shared across every sub-request of a single chunked call via the `_chunked_client` `ContextVar` so the paginated helpers (`_walk_pages`) reuse it.

## Retries (`API_USGS_RETRIES`)

Each sub-request is retried on a transient failure with exponential backoff + full jitter, honoring a server `Retry-After`. A long `Retry-After` (multi-minute quota window) escalates immediately to the resumable interruption rather than blocking the call.

| Value | Behavior |
|---|---|
| unset / blank | retry, cap = **4** (default) |
| `N ≥ 1` | up to `N` retries per sub-request |
| `0` | no retry — first transient surfaces as a resumable `ChunkInterrupted` |
| negative, malformed | `ValueError` at call time |

- **`RetryPolicy`** — frozen value object (`from_env` / `should_retry` / `backoff`). Full jitter de-correlates concurrent retries so a throttle event doesn't make them re-burst in lockstep.
- **`_retryable`** is narrower than `_classify_chunk_error`: it retries the typed `RateLimited` / `ServiceUnavailable` / `httpx.TransportError` but **not** `httpx.InvalidURL` (a too-long cursor won't fix on retry).
- Retries fire **inside the pool slot**, so a backing-off chunk holds its connection — effective concurrency shrinks gracefully under throttling.

## Architecture

**`ChunkPlan`** — extracts every chunkable axis (multi-value list params + the cql-text `filter` along its top-level `OR` clauses), greedy-halves the biggest axis until each sub-request URL fits the byte budget, and iterates the cartesian product. Single-request queries get a one-step plan that skips chunking entirely.

**`ChunkedCall`** — drives the plan over an `async def fetch(args)`:
- `_run` gathers every pending sub-request under one shared `httpx.AsyncClient`.
- `resume()` is the sync facade — runs `_run` through an `anyio` blocking portal so it works whether or not the caller is already in an event loop (Jupyter / IPython / async apps).
- `partial_frame` / `partial_response` snapshot completed sub-requests as raw `(frame, response)` — safe to inspect from inside a `ChunkInterrupted` handler without triggering `finalize` (which for OGC getters includes a schema fetch on an empty frame).
- `finalize` hook (dependency-injected callable) lets the OGC getters inject `utils._finalize_ogc` (type coercion, column arrangement, `BaseMetadata`, `max_rows` truncation) without coupling the chunker to OGC concerns. A successful first call and a resumed call yield the same shape.

**Failure precedence in the gather** (in order):
1. **Cancellation signals** (`CancelledError`, `KeyboardInterrupt`, `SystemExit`) propagate unmodified — wrapping them would consume the user's stop signal.
2. **Non-transient failures** (real bugs unrecognized by `_classify_chunk_error`) surface raw, so they aren't masked behind a resumable handle for a transient sibling that landed later.
3. **Recognized transients** raise the first one as a resumable `ChunkInterrupted` subclass (`QuotaExhausted` for 429, `ServiceInterrupted` for the rest).

## Deps

- `geopandas>=0.10` + `mapclassify` added to `[doc]` extras so `WaterData_demo.ipynb`'s `.set_crs().explore()` cell executes (the plain-pandas frame lacks `.set_crs`).

## Validation

| | Result |
|---|---|
| **Offline test suite** | 298 passed, ruff clean |
| **Live API regression suite** (`tests/waterdata_test.py`) | 63 passed |
| **README examples** | All 4 byte-identical to `main` (matching row counts, columns, SHA-256 of head rows — across single-site, multi-site, POST/CQL2 1,360-row, and 52,853-row paginated continuous queries) |
| **Notebook execution** (8 from `demos/`) | 7/7 OK on both branches |
| **Cell-by-cell notebook diff** | All data cells match between branches |

## Out of scope (intentional)

- **Adaptive concurrency (AIMD)** — a client-side token-bucket controller that converges to the gateway's steady-state rate would reduce retries further. The pool-held-during-backoff behavior is a partial version; a full controller is a separate change.
- **`asyncio.TaskGroup` (3.11+)** — would replace `gather` once the 3.9 floor is dropped, though the partial-completion contract fights `TaskGroup`'s cancel-on-first-failure default.
- **Pre-emptive quota gating** — an earlier revision raised a `RequestExceedsQuota` before a burst the `x-ratelimit-remaining` header said couldn't fit; dropped as not worth the complexity. The reactive retry-then-escalate flow is the chosen alternative.
- **Layered config refactor** — gathering the four runtime knobs behind a `WaterDataConfig` dataclass with file + env + Python-override precedence is in progress as a follow-up (#3 on the fork).


dataretrieval/waterdata/chunking.py
```diff
@@ -4,19 +4,40 @@
...
+for each axis that minimizes total sub-requests while keeping every
+sub-request URL under the budget; ``ChunkedCall`` fetches the resulting
+cartesian product of chunks. Requests that already fit get a trivial
 single-step plan — ``ChunkedCall`` has one code path either way.
 
+Concurrency: ``multi_value_chunked`` fans every pending sub-request out
+under one ``asyncio.gather`` sharing a single ``httpx.AsyncClient``;
+concurrency is bounded purely by the client's connection pool
+(``httpx.Limits(max_connections=N, max_keepalive_connections=N)``), so
+the pool — not a semaphore — throttles. ``API_USGS_CONCURRENT`` resolves
```
thodson-usgs: do we need to clarify that this is not a sephamore?

dataretrieval/waterdata/chunking.py
```diff
@@ -4,19 +4,40 @@
...
+``N``: an integer N > 1 caps connections at N; ``1`` pins a single
+connection (one request at a time); the literal ``unbounded`` removes
+the cap (``N=None``). The default (16) is the server-friendly sweet
+spot; higher values can trip USGS burst-protection 5xx in practice. The
```
thodson-usgs, line 20: The default is N=16; higher values can trigger back off errors.  Lets test 32 and 64. Lower the url limit in order to boost the chunk count in the tests

dataretrieval/waterdata/chunking.py
```diff
@@ -93,22 +109,236 @@
...
+    if value < 1:
+        raise ValueError(
+            f"{_CONCURRENCY_ENV} must be >= 1 (got {value}); use "
+            f"'{_CONCURRENCY_UNBOUNDED}' to disable the cap."
+        )
+    return value
+
+
+# Retry-with-backoff defaults for transient sub-request failures (429 /
+# 5xx / connect-read timeouts). All four are resolved at call time by
```
thodson-usgs: suggest you omit "All four are" ... through "power users"

dataretrieval/waterdata/chunking.py
```diff
@@ -93,22 +109,236 @@
...
+        raise ValueError(f"{_RETRIES_ENV} must be >= 0 (got {value}).")
+    return value
+
+
+@dataclass(frozen=True)
+class RetryPolicy:
+    """Bounded retry-with-backoff config for transient sub-request failures.
+
+    An immutable value object that owns the *timing* decisions; the
+    exception taxonomy ("is this worth retrying at all?") lives in
```
thodson-usgs: I this a comment to yourself? ("is this worth retrying at all?")? If so remove it or implement the fix.

dataretrieval/waterdata/chunking.py
```diff
@@ -140,7 +372,26 @@ def get_active_client() -> httpx.Client | None:
...
+_Fetch = Callable[[dict[str, Any]], Awaitable[tuple[pd.DataFrame, httpx.Response]]]
+
+# Caller-supplied transform applied to the *combined* chunk result. It lets a
+# resumed call (:meth:`ChunkedCall.resume` / :attr:`~ChunkedCall.partial_frame`
+# / :attr:`~ChunkedCall.partial_response`) return the same shape as the
+# un-interrupted call instead of the chunker's raw ``(frame, httpx.Response)``.
+# The chunker stays generic — it only knows "post-process the assembled
+# result"; the OGC getters inject the actual type-coercion / column-arrangement
+# / ``BaseMetadata`` pipeline (see ``utils._finalize_ogc``). The default is
+# identity, so direct ``ChunkedCall`` use and the tests are unaffected.
```
thodson-usgs: This is a little dense. Please copy edit to improve readability.

dataretrieval/waterdata/chunking.py
```diff
@@ -850,6 +1110,118 @@ def _classify_chunk_error(
...
+    the ``__cause__`` chain for resumability. ``_paginate`` raises an
+    initial-request transient (429 / 5xx / :class:`httpx.TransportError`
+    such as ``ConnectError`` / ``ReadTimeout``) *raw*, but re-wraps any
+    mid-pagination failure as a ``RuntimeError``. Retrying only the raw,
+    top-level transient means we re-issue a sub-request that made no
+    progress (cheap), while a failure after partial pagination escalates
+    to the resumable :class:`ChunkInterrupted` instead of being re-walked
+    from page 1 — which would re-spend the very quota that was exhausted.
+    ``httpx.InvalidURL`` is excluded (a too-long cursor won't fix on
+    retry), and it only ever arises on a follow-up page anyway.
```
thodson-usgs: A little too dense. Please copy edit to improve readability

dataretrieval/waterdata/chunking.py
```diff
@@ -1097,69 +1614,134 @@ def resume(self) -> tuple[pd.DataFrame, httpx.Response]:
...
-        resumable :class:`ChunkInterrupted` carrying this call, or
-        re-raise it unchanged to preserve its type. Catches
-        ``RuntimeError`` (the layer's typed contract:
-        :class:`RateLimited`, :class:`ServiceUnavailable`, or the
-        mid-pagination wrapper), :class:`httpx.HTTPError`
-        (transport-level failures like ``ConnectError`` /
-        ``TimeoutException``), and :class:`httpx.InvalidURL` (which
-        inherits directly from ``Exception``, not ``HTTPError``); all
-        three feed :func:`_classify_chunk_error`.
+        The async execution core: gather every pending sub-request over
```
thodson-usgs: Maybe omit "The async execution core" I'll leave it up to you.

dataretrieval/waterdata/chunking.py
```diff
@@ -4,19 +4,41 @@
...
+portal), so it works whether or not the caller is already inside an
+event loop (Jupyter / IPython / async apps).
+
+Retries: each sub-request is retried on a transient failure (429,
+5xx, connect/read timeout) with exponential backoff + full jitter,
+honoring a server ``Retry-After`` when present. ``API_USGS_RETRIES``
+sets the cap (default 4; ``0`` disables). A ``Retry-After`` longer
+than the per-call ceiling isn't slept off inline — it escalates to
+the resumable interruption below so a multi-minute quota-window
+reset doesn't block the call.
```
thodson-usgs:
```suggestion
than the per-call ceiling escalates to a resumable interruption.
```

dataretrieval/waterdata/chunking.py
```diff
@@ -140,7 +365,24 @@ def get_active_client() -> httpx.Client | None:
...
+# column arrangement, ``BaseMetadata``) through ``utils._finalize_ogc``.
+# The default is identity, so direct ``ChunkedCall`` use is unaffected.
+_Finalize = Callable[[pd.DataFrame, httpx.Response], tuple[pd.DataFrame, Any]]
+
+
+def _passthrough_result(
+    frame: pd.DataFrame, response: httpx.Response
+) -> tuple[pd.DataFrame, Any]:
+    """Default :data:`_Finalize`: return the raw combined pair unchanged."""
+    return frame, response
```
thodson-usgs, line 383: Is this function useful? Does it result in cleaner more readable code?
Furthermore, is this always applied to the Finalize callable in practice? If so do we really need Finalize as a parameter?

dataretrieval/waterdata/chunking.py
```diff
@@ -850,6 +1100,115 @@ def _classify_chunk_error(
...
+        (seconds) when the transient carried one, else ``None``.
+    """
+    if isinstance(exc, (RateLimited, ServiceUnavailable)):
+        return True, exc.retry_after
+    if isinstance(exc, httpx.TransportError):
+        return True, None
+    return False, None
+
+
+# Sleep hook, indirected through a module global so tests can
```
thodson-usgs: is there a more pythonic way to accomplish this? Rather then setting a global in the middle of the module?

dataretrieval/waterdata/chunking.py
```diff
@@ -850,6 +1100,115 @@ def _classify_chunk_error(
...
+    -------
+    tuple[bool, float or None]
+        ``(retryable, retry_after)`` — the server ``Retry-After`` hint
+        (seconds) when the transient carried one, else ``None``.
+    """
+    if isinstance(exc, (RateLimited, ServiceUnavailable)):
+        return True, exc.retry_after
+    if isinstance(exc, httpx.TransportError):
+        return True, None
+    return False, None
```
thodson-usgs, line 1126: This is a trivial function and it's called in exactly one place. If you agree, then I think we can consolidate this.

dataretrieval/waterdata/chunking.py
```diff
@@ -767,26 +1006,37 @@ def iter_sub_args(self) -> Iterator[dict[str, Any]]:
...
-            Function that issues a single sub-request, given the
+        fetch : Callable
+            ``async def`` that issues a single sub-request, given the
             substituted args dict, and returns ``(frame, response)``.
+        retry_policy : RetryPolicy, optional
+            Per-sub-request retry-with-backoff policy. Defaults to
+            :data:`_NO_RETRY`; the decorator passes ``RetryPolicy.from_env()``.
+        finalize : Callable, optional
+            Transform applied to the combined ``(frame, response)`` (see
+            :data:`_Finalize`). Defaults to :func:`_passthrough_result`.
```
thodson-usgs, line 1029: See my earlier point. If finalize always defaults to _passthrough_result in our code, then maybe we don't need it as a parameter.

dataretrieval/waterdata/chunking.py
```diff
@@ -977,62 +1336,185 @@ class ChunkedCall:
...
         The plan being driven (read-only after construction).
-    fetch_once : Callable
-        The per-sub-request fetch function.
+    fetch : Callable
+        The async per-sub-request fetch function.
+    finalize : Callable
+        Transform applied to the combined result (see :data:`_Finalize`) at
+        the terminal :meth:`_run` return, so a completed call yields the
+        caller's finished shape. The ``partial_*`` accessors deliberately
+        skip it and stay raw.
```
thodson-usgs, line 1374: same point as other finalize.

dataretrieval/waterdata/chunking.py
```diff
@@ -977,62 +1336,185 @@ class ChunkedCall:
...
+        (see :meth:`_combine_raw`).
+
+        Parameters
+        ----------
+        index : int
+            The sub-args index this completed pair belongs to.
+        pair : tuple of (pandas.DataFrame, httpx.Response)
+            The completed sub-request's ``(frame, response)`` pair.
+        """
+        self._chunks[index] = pair
```
thodson-usgs: Is this function useful? Is the code clear and simple if we omit it? If so, then I think we should remove it.

dataretrieval/waterdata/chunking.py
```diff
@@ -977,62 +1336,185 @@ class ChunkedCall:
...
+
+        Returns
+        -------
+        tuple of (pandas.DataFrame, finalized response)
+            The combined frame and whatever :attr:`finalize` produces —
+            a raw :class:`httpx.Response` by default, or the OGC
+            getters' type-coerced / column-arranged frame plus
+            ``BaseMetadata``.
+        """
+        return self.finalize(*self._combine_raw())
```
thodson-usgs: Here's another trivial one-line function. Consider whether this is actually useful. If its only called once, i think it would be a strong candidate for removal.

thodson-usgs (commented):
please address my comments.

dataretrieval/waterdata/chunking.py
thodson-usgs: Checked the actual flow before changing this — `finalize` is **never** the default in production. Every OGC getter goes through `utils.get_ogc_data`, which builds a `functools.partial(_finalize_ogc, properties=..., output_id=..., service=..., max_rows=...)` per call and passes it as `finalize=` to `_fetch_once` (`utils.py:1387–1396`). So `_passthrough_result` is the fallback for direct chunker use (tests, future non-OGC callers), not the common path.

The injection is what keeps `chunking.py` from importing OGC-specific post-processing. Without it, the chunker would either need to import `_finalize_ogc` directly (couples chunking → utils) or `resume()` would have to return the *raw* shape — and callers catching `ChunkInterrupted` further up the stack would then have to know to apply OGC post-processing themselves (breaks the "`resume()` returns the same shape a successful first call would have" contract).

I'd push back on removing it. Happy to swap the named `_passthrough_result` for a `lambda f, r: (f, r)` if that small abstraction feels like overkill — but the parameter itself earns its place.

dataretrieval/waterdata/chunking.py
thodson-usgs: See [my reply on the `_passthrough_result` thread](https://github.com/DOI-USGS/dataretrieval-python/pull/285#discussion_r3318236908) — `finalize` is never actually defaulted in production; every OGC getter overrides it with a `functools.partial(_finalize_ogc, ...)`. The default exists as a fallback, not the common path.

dataretrieval/waterdata/chunking.py
thodson-usgs: Same as the other `finalize` threads — see [the main reply](https://github.com/DOI-USGS/dataretrieval-python/pull/285#discussion_r3318236908). `finalize` is always overridden in production by the OGC getters, so the parameter earns its place.

dataretrieval/waterdata/chunking.py
thodson-usgs: Declining this one — `_retryable` is **directly unit-tested** (`tests/waterdata_chunking_test.py:1571–1591`, 7 assertions covering the `RateLimited` / `ServiceUnavailable` / `ReadTimeout` / `ConnectError` / `InvalidURL` / `RuntimeError` cases and the `__cause__`-walking semantics). Inlining it into `_retry_delay` would either delete that coverage or force contorted "call `_retry_delay` and assert it returned `None`" assertions that lose the per-case specificity.

Keeping `_retryable` as a named, separately-testable concept earns the indirection. Open to revisiting if you have a different test pattern in mind.

dataretrieval/waterdata/chunking.py
thodson-usgs: Belated answer with the data (thread is already resolved — posting for the record):

Stress-tested 16 / 32 / 64 against the live API. Lowered `_WATERDATA_URL_BYTE_LIMIT` to fan a single `get_daily` out to ~64 sub-requests over 400 real WI stream sites, instrumented sub-request and retry counts.

**Burst test (equal 64-request volume, varied simultaneity):**

| concurrency | trials | outcome | retries fired |
|---|---|---|---|
| 16 | 3/3 | all OK | 0 |
| 32 | 3/3 | all OK | 0 |
| 64 | 3/3 | all OK | 0 |

So 32/64 are reliable concurrency-wise — no per-connection burst-protection 5xx.

**The actual ceiling is the hourly request quota, not concurrency.** After ~600 cumulative requests in ~2 min, calls began returning `QuotaExhausted` (429) persistently across *all* levels. This is concurrency-independent — same volume at conc=16 would exhaust it too. Higher concurrency just reaches the quota faster.

The retry escalation path is correct: `retries=4` runs hit `QuotaExhausted` with **0 retries fired** — the 429's `Retry-After` exceeds the 60s cap, so it escalates to a resumable `QuotaExhausted` rather than blocking inline.

Recommendation: keep the default at 16 (as you said), but the docstring claim "higher values can trip USGS burst-protection 5xx" wasn't observed at 64 — the only backoff was hourly-quota 429. Happy to tighten the wording if you'd like.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — dropped the "not a semaphore" aside.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — dropped that sentence.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — dropped the parenthetical note-to-self.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — copy-edited the `_Finalize` comment for readability.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — copy-edited the `_retryable` docstring for readability.

dataretrieval/waterdata/chunking.py
thodson-usgs: Done in bbabf50 — dropped the "The async execution core" lead-in.

dataretrieval/waterdata/chunking.py
thodson-usgs: Applied as suggested in c63da1b.

dataretrieval/waterdata/chunking.py
thodson-usgs: Agreed — dropped `_ASLEEP` in c63da1b. Production now uses `asyncio.sleep` directly; tests patch `chunking.asyncio.sleep`.

dataretrieval/waterdata/chunking.py
thodson-usgs: Agreed — `record()` inlined in c63da1b. The "single writer of `_chunks`" invariant moves to a comment on the dict initialization.

dataretrieval/waterdata/chunking.py
thodson-usgs: Agreed — `combined()` inlined in c63da1b. The "`partial_*` bypass" note moves to a comment at the return site, where its more useful than buried in a removed helper.

PR #289, opened 2026-05-25
feat(http): Migrate from requests to httpx

thodson-usgs:
## Summary

- Swaps the HTTP client across the package: `requests.Session` → `httpx.Client`, `requests.Response` → `httpx.Response`. Same call shape; same response shape; only the underlying transport changes.
- Centralizes header / timeout / redirect defaults in `dataretrieval.utils.HTTPX_DEFAULTS` (`httpx.Timeout(60.0, connect=10.0)`, `follow_redirects=True`).
- Routes the chunker, paginated-loop helpers, and OGC waterdata fetchers through one `httpx.Client` that the chunker publishes on the `_chunked_client` ContextVar so paginated sub-requests reuse the connection pool across a single chunked call.
- Migrates the test suite from `requests_mock` to native `pytest-httpx`; the old `_RequestsMockShim` is gone.

## Why httpx

`httpx` ships sync and async clients on a **unified API**, so the same request shape can power both the existing synchronous getters and a follow-up async parallel chunker (see #285, stacked on this PR). `requests` is still actively maintained but has no first-class async story — adding one for #285 under `requests` would have been a `ThreadPoolExecutor` bolt-on, one-off, and not reusable.

Other minor wins: end-to-end type hints, HTTP/2 support if we ever want it, smaller wheel footprint vs `requests + chardet`.

## Backwards-compat surface

- `BaseMetadata.header` is now `httpx.Headers` instead of `requests.structures.CaseInsensitiveDict`. Case-insensitive `.get(...)` still works; literal dict equality (`md.header == {"k": "v"}`) no longer holds because `httpx.Headers` carries auto-added entries.
- `BaseMetadata.url` is coerced to `str` (was already `str`-ish; now explicit `str(httpx.Response.url)`).
- `RequestExceedsQuota` and `API_USGS_LIMIT` are removed (per review). The chunker no longer pre-empts on `x-ratelimit-remaining`; a natural 429 still surfaces as `QuotaExhausted` carrying `.call` for `.call.resume()` recovery.

## Three httpx behavior diffs handled defensively

- `httpx.InvalidURL` is raised client-side when a URL component > 64 KB. Caught by `_safe_request_bytes` (treats "too big to construct" as "doesn't fit", so the planner's halving loop keeps shrinking) AND by `_issue` / `_classify_chunk_error` (treats a runtime InvalidURL as `ServiceInterrupted` so partial state remains recoverable via `.call.resume()`). Note: `httpx.InvalidURL` does NOT inherit from `httpx.HTTPError` — needs an explicit catch.
- `httpx.Response.elapsed` is only populated when the response is closed, and `pytest-httpx` mock responses don't populate it. The new `_safe_elapsed` helper falls back to `timedelta(0)`.
- `httpx.Response.url` is a read-only property. The new `_set_response_url` helper rewrites it by reseating the bound request, with a fallback path for `Mock`-shaped test responses.

## Test plan

- [x] **404 mocked tests** pass after migrating to `pytest-httpx`. The new `tests/conftest.py` is ~30 lines configuring pytest-httpx strict-mode relaxations.
- [x] New regression tests:
  - `test_connection_error_wrapped_as_service_interrupted` — `httpx.ConnectError` mid-fetch surfaces as resumable `ServiceInterrupted`
  - `test_invalid_url_wrapped_as_service_interrupted` — `httpx.InvalidURL` mid-fetch (e.g., oversize cursor URL) surfaces as resumable `ServiceInterrupted`
  - `test_query_nldi_non_200_surfaces_reason_phrase` — locks in the `.reason` → `.reason_phrase` migration miss
- [x] `ruff check` and `ruff format --check` pass.
- [x] CI green (lint, docs, ubuntu/windows × 3.9/3.13/3.14).

## Stacked PR

The async parallel chunker (#285) sits on top of this branch and uses the same `httpx.AsyncClient`. After this merges, #285 rebases onto `main` for a clean diff.

### DIFF
--- a/dataretrieval/utils.py
+++ b/dataretrieval/utils.py
@@ -5,11 +5,16 @@ Useful utilities for data munging.
 import warnings
 from collections.abc import Iterable
 
+import httpx
 import pandas as pd
-import requests
 
 import dataretrieval
 from dataretrieval.codes import tz
+
+HTTPX_DEFAULTS = {
+    "follow_redirects": True,
+    "timeout": httpx.Timeout(60.0, connect=10.0),
+}
 
 
 def to_str(listlike, delimiter=","):
@@ -205,7 +210,7 @@ class BaseMetadata:
         Response url
     query_time: datetme.timedelta
         Response elapsed time
-    header: requests.structures.CaseInsensitiveDict
+    header: httpx.Headers
         Response headers
 
     """
@@ -216,7 +221,7 @@ class BaseMetadata:
         Parameters
         ----------
         response: Response
-            Response object from requests module
+            Response object from httpx module
 
         Returns
         -------
@@ -225,8 +230,8 @@ class BaseMetadata:
 
         """
 
-        # These are built from the API response
-        self.url = response.url
+        # Coerce httpx.URL -> str: BaseMetadata.url has always been str.
+        self.url = str(response.url)
         self.query_time = response.elapsed
         self.header = response.headers
         self.comment = None
@@ -254,10 +259,29 @@ class BaseMetadata:
         return f"{type(self).__name__}(url={self.url})"
 
 
+_URL_TOO_LONG_EXAMPLE = """
+                    # n is the number of chunks to divide the query into \n
+                    split_list = np.array_split(site_list, n)
+                    data_list = []  # list to store chunk results in \n
+                    # loop through chunks and make requests \n
+                    for site_list in split_list: \n
+                        data = nwis.get_record(sites=site_list, service='dv', \n
+                                               start=start, end=end) \n
+                        data_list.append(data)  # append results to list"""
+
+
+def _url_too_long_error(detail: str) -> ValueError:
+    return ValueError(
+        "Request URL too long. Modify your query to use fewer sites. "
+        f"{detail}. Pseudo-code example of how to split your query: "
+        f"\n {_URL_TOO_LONG_EXAMPLE}"
+    )
+
+
 def query(url, payload, delimiter=",", ssl_check=True):
     """Send a query.
 
-    Wrapper for requests.get that handles errors, converts listed
+    Wrapper for httpx.get that handles errors, converts listed
     query parameters to comma separated strings, and returns response.
 
     Parameters
@@ -265,7 +289,7 @@ def query(url, payload, delimiter=",", ssl_check=True):
     url: string
         URL to query
     payload: dict
-        query parameters passed to ``requests.get``
+        query parameters passed to ``httpx.get``
     delimiter: string
         delimiter to use with lists
     ssl_check: bool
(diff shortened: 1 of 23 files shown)


dataretrieval/utils.py
```diff
@@ -5,12 +5,17 @@
...
+import httpx
 import pandas as pd
-import requests
 
 import dataretrieval
 from dataretrieval.codes import tz
 
+# Shared kwargs for every ``httpx`` call site. ``requests`` defaulted to
+# following redirects on GET; ``httpx`` does not. ``timeout=None``
+# preserves the no-timeout behavior the chunker docs promise.
```
thodson-usgs: Omit this comment. Just determine the correct default behavior

dataretrieval/utils.py
```diff
@@ -299,24 +328,14 @@ def query(url, payload, delimiter=",", ssl_check=True):
...
-                    for site_list in split_list: \n
-                        data = nwis.get_record(sites=site_list, service='dv', \n
-                                               start=start, end=end) \n
-                        data_list.append(data)  # append results to list"""
         raise ValueError(
             "Request URL too long. Modify your query to use fewer sites. "
-            + f"API response reason: {_reason}. Pseudo-code example of how to "
-            + f"split your query: \n {_example}"
+            f"API response reason: {response.reason_phrase}. Pseudo-code "
+            f"example of how to split your query: \n {_URL_TOO_LONG_EXAMPLE}"
```
thodson-usgs: Why are we duplicating this error message? form lines 316 to 319?

dataretrieval/waterdata/chunking.py
```diff
@@ -9,18 +9,18 @@
...
-Quota: after the first sub-request ``ChunkedCall`` reads
-``x-ratelimit-remaining``; if the rest of the plan won't fit, it
-raises ``RequestExceedsQuota`` before burning more budget. Set
-``API_USGS_LIMIT=0`` to skip this pre-emptive check and attempt the
-full plan anyway.
+Quota: each completed sub-request's ``x-ratelimit-remaining`` is
+checked against the still-pending count, and
+``RequestExceedsQuota`` fires before the next sub-request if the
+window can't cover the rest. Set ``API_USGS_LIMIT=0`` to skip this
+pre-emptive check and attempt the full plan anyway.
```
thodson-usgs: I think we can remove this error. We don't really have a good idea of how many requests will be needed due to pagination.

dataretrieval/waterdata/chunking.py
```diff
@@ -98,42 +100,42 @@
...
-_chunked_session: ContextVar[requests.Session | None] = ContextVar(
+# client in that case.
+_chunked_session: ContextVar[httpx.Client | None] = ContextVar(
     "_chunked_session", default=None
 )
 
 
 @contextmanager
-def _publish_session(session: requests.Session) -> Iterator[None]:
+def _publish_session(client: httpx.Client) -> Iterator[None]:
```
thodson-usgs: many private functions using the term "session". Should these be updated to "client" as per httpx terminology?

dataretrieval/waterdata/chunking.py
```diff
@@ -607,12 +675,14 @@ class ChunkPlan:
...
-        URL there, so ``ChunkedCall`` skips the override to avoid an
-        extra ``build_request`` call on the hot path.
+        URL of the user's original (un-chunked) request, used to
+        overwrite a chunked response's ``.url`` so ``BaseMetadata``
+        reflects the full query. Set to the original-input URL when
+        chunking is needed; falls back to a worst-case sub-request
+        URL when even the original URL trips httpx's 64 KB cap.
+        ``None`` on the nothing-to-chunk passthrough path
+        (``fetch_once``'s response already carries the canonical URL)
+        and when no buildable URL exists.
```
thodson-usgs: This doc is too long. Can you reduce canonical_url doc to 3-4 lines.

dataretrieval/waterdata/chunking.py
```diff
@@ -1123,59 +1248,62 @@ def resume(self) -> tuple[pd.DataFrame, requests.Response]:
             is on ``exc.call`` — wait for the underlying condition to
             clear and call ``exc.call.resume()`` again.
         RequestExceedsQuota
-            When the rate-limit window can't cover the remaining plan
-            (checked after the first sub-request).
+            When the rate-limit window can't cover the still-pending
+            plan (checked after every non-final sub-request).
```
thodson-usgs: Omit this. We're not going to check. We'll just let it fail once the quota is reached.

thodson-usgs (commented):
Please address my comments.

PR #318, opened 2026-06-01
refactor(waterdata): decompose the 2033-LOC utils.py god-module into cohesive modules

thodson-usgs:
> **Rewritten and rebased onto #324.** The original PR split `waterdata/utils.py`; #324 has since made that file thin (it extracted the generic engine into `ogc/`). This PR retargets #318's original intent — *decompose the OGC god-module by cohesive domain* — onto the post-#324 layout. **Stacked on #324: review/merge after it lands.** Until #324 merges, the diff below includes #324's commits; it collapses to just the engine decomposition once #324 is in `main`.

## Why

After #324, the god-module is `dataretrieval/ogc/engine.py` (~1940 LOC) — it absorbed several unrelated concerns (constants/config, per-call request context, HTTP error handling, arg validation, request building, response shaping) on top of the async pagination driver. Changing how requests are built means navigating the same file as datetime parsing, type coercion, and the async driver.

## What

Split it into cohesive **private** sibling modules under `dataretrieval/ogc/`, moving every definition **AST-identically** (no signature/logic change):

| module | holds |
|---|---|
| `_constants.py` | URLs, `OgcDialect`, regexes, param sets, the geopandas probe (pure-data leaf) |
| `_context.py` | per-call request context — the base-url / dialect / row-cap context vars + their managers |
| `_http.py` | headers, `_error_body`, `_raise_for_non_200`, retry-after |
| `_validate.py` | **pure** arg normalization/validation + id switching |
| `_requests.py` | request building (GET / CQL2, date formatting, pagination URLs, the queryables probe) |
| `_responses.py` | the whole wire-response → DataFrame domain (parse + shape + finalize) |
| `engine.py` | async pagination driver + `get_ogc_data`; re-exports the above |

`engine.py` (**1937 → 570 LOC**) stays the public **façade**: it re-exports every name (explicit `__all__`) so every existing `from dataretrieval.ogc.engine import …` site in `waterdata`, `ngwmn`, and the tests keeps working — **no import sites touched**. The dependency DAG is acyclic and leaf-first: `_constants ← _context ← {_http, _validate, _requests, _responses} ← engine`. `_validate` is a genuine pure leaf — the one live HTTP probe (`_check_ogc_requests`) lives in `_requests` with the other request builders.

## Behavior-preserving

- All **42** top-level symbols moved with **byte-identical AST bodies** (verified mechanically vs the pre-split file) — none lost, changed, or duplicated.
- **One** test change: the single `gpd` monkeypatch seam was relocated from `engine` to `_responses` (where the geopandas parse functions now live), so the response domain coheres by *domain* rather than by a patch target.
- **296 offline tests pass**; `ruff` + `mypy --strict` clean; no import cycles.

## Scope

`ogc/chunking.py` (1776 LOC) was evaluated and **deliberately left intact** — unlike `engine.py`'s unrelated concerns, it is one cohesive algorithm orbiting `ChunkedCall` (the exception taxonomy, retry policy, and executor share state and a cross-cutting test seam), so splitting would fragment it and add a mandatory façade for little gain.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,17 @@ repos:
         args: [--fix]
       - id: ruff-format
 
+  # Type-check with mypy --strict (config lives in pyproject [tool.mypy]).
+  # Pinned to the same major as CI's `mypy<2`. httpx/anyio are installed into
+  # the isolated hook env so their types resolve — without them mypy falls back
+  # to `Any` and mis-reports (the runtime deps aren't in the hook's venv).
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.20.2
+    hooks:
+      - id: mypy
+        pass_filenames: false
+        additional_dependencies: [httpx, anyio]
+
   # Strip cell outputs + execution_count from notebooks on commit so the
   # diff is the source, not the rendered run. Demos still execute fine
   # locally; clean commits keep PRs reviewable and avoid quota/timestamp
--- a/README.md
+++ b/README.md
@@ -10,8 +10,8 @@
 Like the original R version
 [`dataRetrieval`](https://github.com/DOI-USGS/dataRetrieval), it retrieves major
 U.S. Geological Survey (USGS) hydrology data types available on the Web, as well
-as data from the Water Quality Portal (WQP) and Network Linked Data Index
-(NLDI).
+as data from the Water Quality Portal (WQP), the National Ground-Water
+Monitoring Network (NGWMN), and the Network Linked Data Index (NLDI).
 
 Check the [NEWS](NEWS.md) for all updates and announcements.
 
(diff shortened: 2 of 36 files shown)


thodson-usgs:
Closing as overtaken by #324. This branch's substantive value — extracting a shared, API-agnostic OGC engine and adding the NGWMN getters as an `ogc` sibling — already merged via #324 (`963eb280`), along with the ChunkInterrupted/resume docs. The only part not in main is the final commit (`32409537`) that splits `ogc/engine.py` into per-concern sub-modules, and it was written against the pre-#324 engine.

After #324, `engine.py` is ~1188 LOC (down from ~1940), so the remaining work is a cosmetic decomposition of a much smaller file that would need full re-authoring against the reshaped engine. Not worth carrying as a stale, conflicting branch. If the engine grows enough to warrant splitting again, we can refile a fresh PR with module boundaries drawn against the current code.

PR #319, opened 2026-06-03
refactor(errors)!: a lean, idiomatic DataRetrievalError taxonomy

thodson-usgs:
## What

A small, idiomatic exception taxonomy: every request failure raises a subclass of `DataRetrievalError`, so one `except dataretrieval.DataRetrievalError` handles any of them. It adds only what the underlying `httpx` exceptions can't express.

```
DataRetrievalError(Exception)         # .status_code / .retry_after / .retryable
├─ HTTPError                  # .status_code — the server returned an error status
│   └─ TransientError         # .retry_after — retryable (429 / 5xx)
│       ├─ RateLimited        #   429
│       └─ ServiceUnavailable #   5xx
├─ RequestTooLarge            # the request can't fit
│   ├─ URLTooLong             #   414 / client-side over-long URL
│   └─ Unchunkable            #   the Water Data chunker can't split the call
├─ NetworkError               # no response: timeout / DNS / refused connection
└─ NoSitesError               # no-data on the legacy nwis path (see "no-data" below)
```

A single factory, `dataretrieval.exceptions.error_for_status(...)`, maps a status to its type, and **every request path routes through it** — the legacy `query` path (`nwis`/`wqp`/`nldi`), the Water Data chunker, and `nadp`/`streamstats` — so a given status surfaces as the same type everywhere.

- **Fatal 4xx** → a generic `HTTPError` carrying `.status_code` (inspect the code rather than a class per code).
- **Retryable** (429 / 5xx) → `RateLimited` / `ServiceUnavailable` (both `TransientError`), carrying `.retry_after`.
- **Domain failures** → `Unchunkable`, `URLTooLong` (keeps the actionable "split your query" message on the legacy path).
- **Connection-level failures** (timeout, DNS, refused) → `NetworkError`, with the underlying `httpx` exception on `__cause__`.

### Branch or retry without knowing the concrete type

Every `DataRetrievalError` exposes three read-anywhere fields — no `isinstance`, no `httpx` import:

- `.status_code` — the HTTP status, or `None` when the failure carried no response.
- `.retry_after` — seconds the server hinted (its `Retry-After`), or `None`.
- `.retryable` — `True` for the transient statuses (429 / 5xx) **and** connection failures; `False` otherwise.

```python
for attempt in range(5):
    try:
        df, md = dataretrieval.waterdata.get_daily(...)
        break
    except dataretrieval.DataRetrievalError as e:
        if not e.retryable or attempt == 4:
            raise
        time.sleep(e.retry_after or 2 ** attempt)
```

A new **"Handling errors"** user-guide page documents these recipes (catch-all, retry-with-backoff, branch-on-status, resume a chunked call).

### No-data is not an error

The modern getters (`waterdata`, `wqp`, `nldi`) **return an empty `DataFrame`** when nothing matches — check `df.empty`, don't catch. Only the deprecated `nwis` (waterservices) path still raises `NoSitesError` on no data.

### Cross-process

The typed errors are picklable via the standard `__getstate__`/`__setstate__` protocol, so they propagate cleanly back from multiprocessing / lithops workers. A `ChunkInterrupted` sheds its live `.call` resume handle across a process boundary (keeping the partial frame/response and counts), while in-process callers still get full `exc.call.resume()`.

## Breaking changes

- Request failures raise typed `DataRetrievalError` subclasses instead of bare `ValueError` / `RuntimeError` / `httpx.HTTPStatusError`. The tree roots **only** at `DataRetrievalError(Exception)` — no more builtin mixins, so `except ValueError` / `except RuntimeError` no longer catch them. This now also covers `ChunkInterrupted` (was a `RuntimeError`) and mid-pagination failures (were a bare `RuntimeError`). Catch `DataRetrievalError` (or a subclass).
- **Connection failures are wrapped as `NetworkError`** instead of raw `httpx` exceptions on the single-shot paths — catch `NetworkError` (or `DataRetrievalError`); the httpx exception is on `__cause__`.
- A fatal 4xx is an `HTTPError` (read `.status_code`); there are no per-code exception types.

## Verification

`ruff` clean (pre-commit hooks). `mypy --strict` and the full `pytest` suite are re-run by CI on push. The taxonomy was previously exercised green locally: typed errors round-trip through pickle / deepcopy (including `ChunkInterrupted` with real partial data and `NetworkError`); every type exposes `.status_code`/`.retry_after`/`.retryable` uniformly; async chunked requests resume after a transient interruption (from both a sync caller and inside a running event loop). Adds a `dataretrieval.exceptions` API reference page, a "Handling errors" user guide, and a `NEWS.md` entry.


dataretrieval/waterdata/chunking.py
```diff
@@ -390,6 +390,12 @@ def _passthrough_result(
     return frame, response
 
 
+# ``RuntimeError`` base retained deliberately: the request-error taxonomy in
+# ``exceptions.py`` shed its ``ValueError`` / ``RuntimeError`` mixins, but
+# ``ChunkInterrupted`` predates that and long-standing callers catch it as a
+# ``RuntimeError``; keeping the base preserves their ``except RuntimeError``
+# handlers. It is *also* a ``DataRetrievalError``, so the unified
+# ``except DataRetrievalError`` clause covers it too.
```
thodson-usgs: don't worry about long-standing callers here. We only recently introduced chunking. You can probably omit historical explanation. We want to jump straight to the longterm solution here.

dataretrieval/waterdata/chunking.py
```diff
@@ -496,6 +502,15 @@ def __init__(
             self.partial_frame = call.partial_frame.copy()
             self.partial_response = call.partial_response
 
+    def _pickle_state(self) -> dict[str, Any]:
+        # ``.call`` is a live ChunkedCall whose ``.fetch`` is a decorated module
+        # function stdlib pickle can't serialize by name, so a real
+        # ChunkInterrupted can't otherwise cross a multiprocessing boundary. Ship
+        # the documented degraded ``call=None`` state instead: ``.resume()`` is
+        # gone (un-resumable cross-process regardless), but the counts, retry
+        # hint, and partial_frame / partial_response survive.
```
thodson-usgs: Similarly, the documentation should focus on the current implementation. Not what was done previously

dataretrieval/waterdata/utils.py
```diff
@@ -1101,15 +1099,15 @@ async def _paginate(
...
-        :class:`~dataretrieval.exceptions.RateLimited` /
-        :class:`~dataretrieval.exceptions.ServiceUnavailable`
-        for 429/5xx, otherwise plain ``RuntimeError`` from
-        :func:`_error_body`), on an initial-page parse failure
-        (wrapped via :func:`_paginated_failure_message` with the
-        original exception on ``__cause__``), or any failure on a
-        subsequent page (same wrapping).
+        On an initial-page parse failure (wrapped via
+        :func:`_paginated_failure_message` with the original exception on
+        ``__cause__``), or any failure on a subsequent page (same wrapping).
```
thodson-usgs: Double check, but we might omit RuntimeError, since this PR should implement the longterm solution.

dataretrieval/__init__.py
```diff
@@ -17,8 +17,10 @@
 ``nldi`` requires geopandas (``pip install dataretrieval[nldi]``) and is
 imported on demand: ``from dataretrieval import nldi``.
 
-Every request failure raises a subclass of :class:`dataretrieval.DataRetrievalError`;
-the taxonomy lives in ``dataretrieval.exceptions``.
+When a request gets an HTTP error response it raises a subclass of
+:class:`dataretrieval.DataRetrievalError` (the taxonomy lives in
+``dataretrieval.exceptions``). Connection-level failures (timeouts, DNS) still
```
thodson-usgs: can we simplify the wording here. Please copy edit.

dataretrieval/exceptions.py
```diff
@@ -1,89 +1,202 @@
...
-:class:`RequestTooLarge` (the request can't fit, however it was issued) and
-:class:`TransientError` (a temporary failure worth retrying).
-
-This module deliberately has no third-party dependencies, so any module can
-import it without pulling in pandas/httpx.
+Every service module (``nwis``, ``wqp``, ``nldi``, ``waterdata``, ``nadp``,
+``streamstats``) raises a subclass of :class:`DataRetrievalError` when a request
+fails, so one ``except dataretrieval.DataRetrievalError`` catches them all.
+Connection-level failures (timeouts, DNS, refused connections) still surface as
+``httpx`` exceptions on the single-shot request paths.
```
thodson-usgs: Is it cleaner or simpler to handle these as DataRetrievalError also?

dataretrieval/exceptions.py
```diff
@@ -1,89 +1,202 @@
...
+        # from a multiprocessing / lithops worker. A subclass holding
+        # un-picklable live state (e.g. the chunker's ``ChunkInterrupted.call``)
+        # overrides ``_pickle_state`` to shed it.
+        return (_rebuild_error, (self.__class__, self.args, self._pickle_state()))
+
+    def _pickle_state(self) -> dict[str, Any]:
+        """Instance state to serialize in :meth:`__reduce__`.
+
+        Defaults to ``__dict__``; override to drop attributes that can't pickle.
+        """
```
thodson-usgs: be sure this is absolutely necessary for the implementation. It smells fish.

dataretrieval/exceptions.py
```diff
@@ -99,56 +212,57 @@ class Unchunkable(RequestTooLarge):
...
-        when the header is absent or unparseable).
-    """
-
-    def __init__(self, message: str, *, retry_after: float | None = None) -> None:
-        super().__init__(message)
-        self.retry_after = retry_after
+#: Deprecated alias for :class:`NoDataError`. The original name leaked NWIS-era
+#: "sites" terminology; it is retained so existing ``except NoSitesError``
+#: handlers keep working, and will be removed in a future release.
+NoSitesError = NoDataError
```
thodson-usgs: Go ahead and add a deprecation worning on NoSitesError

dataretrieval/utils.py
```diff
@@ -348,13 +344,11 @@ def query(
...
-        :class:`~dataretrieval.exceptions.NotFoundError` (404),
-        :class:`~dataretrieval.exceptions.URLTooLong` (414 or a client-side
-        over-long URL), :class:`~dataretrieval.exceptions.ServiceUnavailable`
-        (5xx), or :class:`~dataretrieval.exceptions.NoSitesError` (no sites/data
-        matched). The 4xx types are also :class:`ValueError`;
-        ``ServiceUnavailable`` is a :class:`RuntimeError`.
+        On an HTTP error response, the typed subclass for the status (see
+        :func:`dataretrieval.exceptions.error_for_status` for the mapping); or
+        :class:`~dataretrieval.exceptions.NoDataError` when a 200 response
+        reports no data matched. Connection-level failures (timeouts, DNS)
```
thodson-usgs: In general, are you sure we treat nodata as an error? I thought the correct pattern was to return an empty data frame.

thodson-usgs (commented):
Please address my suggestions.

PR #343, opened 2026-07-07
fix(waterdata): attach EPSG:4326 CRS to OGC GeoDataFrames

arpitjain099:
This fixes #342. The waterdata getters build their GeoDataFrame with gpd.GeoDataFrame.from_features(...) but never pass a crs, so the result comes back with .crs of None even though the coordinates are published in EPSG:4326 and the get_monitoring_locations docstring says as much. That leaves the frame CRS-naive, so to_crs and spatial joins raise on an otherwise valid result, and it's out of step with the nldi (EPSG:4326) and nwis (EPSG:4269) modules that already tag theirs.

I added a shared _CRS constant in ogc/shaping.py and pass it at both build sites (shaping and waterdata/stats), mirroring how nldi/nwis do it. I confirmed the fix with new offline regression tests that fail before the change (crs is None) and pass after (crs is EPSG:4326) for both the sites path and the stats path; the rest of the waterdata suite and ruff stay green. Thanks for taking a look.

### DIFF
--- a/dataretrieval/ogc/shaping.py
+++ b/dataretrieval/ogc/shaping.py
@@ -31,6 +31,12 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+# Water Data OGC coordinates are published in WGS84, so attach that CRS where the
+# GeoDataFrame is built (otherwise the result is CRS-naive and ``to_crs`` /
+# spatial joins fail). Mirrors the constants in ``nldi`` (EPSG:4326) and ``nwis``
+# (EPSG:4269).
+_CRS = "EPSG:4326"
+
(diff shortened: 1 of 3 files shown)


thodson-usgs:
thanks @arpitjain099, @matinguku

@mikemahoney218-usgs, before I merge this, could you confirm whether or not EPSG:4326 is unversal. If not, we might assign the crs with a warning.

mikemahoney218-usgs:
Universal unless a user provides the `crs` query argument, in which case we use the CRS requested (currently only supports OGC's variants of NAD83 and NAD27). We mostly don't document that functionality because the way OGC handles CRS is needlessly complex, though, so I'd suspect ~100% of the time it's 4326

PR #345, opened 2026-07-23
perf(ogc): faster page parsing and chunk combination

thodson-usgs:
## Summary

Two client-side performance improvements to the OGC page-parsing and chunk-combination paths, identified through profiling the `parallel_chunks` fan-out code.

### Changes

1. **Replace `pd.json_normalize` with `pd.DataFrame` in `_get_resp_data`** (shaping.py)

   OGC API feature properties are flat dictionaries — `json_normalize`'s recursive nesting support is unnecessary overhead. Direct `pd.DataFrame` construction is **2.4× faster** per page (3.2 ms → 1.35 ms on 2500-feature pages).

2. **Skip `drop_duplicates` when the plan has no filter axis** (planning.py, chunking.py)

   List-axis chunks (the common `parallel_chunks` case — splitting on `monitoring_location_id`, `parameter_code`, etc.) produce non-overlapping partitions by construction. Only filter-axis chunks (splitting on CQL `OR` clauses) can overlap. `_combine_chunk_frames` now accepts a `dedup` flag and `ChunkedCall._combine_raw` passes `dedup=self.plan.has_filter_axis` — **2.7× faster** combine (3.4 ms → 1.25 ms on 52k-row results) for the common case.

### Impact

For a typical 32-chunk query fetching 32 pages (~52k rows):
- Page parse savings: ~1.85 ms × 32 pages = **~59 ms saved**
- Combine savings: **~2.2 ms saved**
- Total: **~61 ms** less client-side overhead per query

These are additive to the network-latency savings `parallel_chunks` already provides.

### Testing

All 574 tests pass. No behavioral change — list-axis chunks were never producing duplicates (the dedup was a no-op); filter-axis chunks still dedup as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ sites, _ = waterdata.get_monitoring_locations(state="Ohio", site_type_code="ST")
 
 with waterdata.parallel_chunks(32):              # fan out into 32 sub-requests
     df, md = waterdata.get_daily(
-        monitoring_location_id=sites["monitoring_location_id"].tolist(),
+        monitoring_location_id=sites["monitoring_location_id"],
         parameter_code="00060",                  # discharge
         time="2004-01-01/2023-12-31",
     )
--- a/README.md
+++ b/README.md
@@ -105,6 +105,58 @@ df, metadata = waterdata.get_continuous(
 print(f"Retrieved {len(df)} continuous gage height measurements")
 ```
 
+#### Speeding up large downloads with `parallel_chunks`
+
+By default the getters split a multi-value request only as far as the server's
+~8 KB URL limit forces — the fewest sub-requests. For a **large, paginated**
+pull that is needlessly conservative: every sub-request pages through its own
+results, so dividing the query into more, smaller sub-requests lets those pages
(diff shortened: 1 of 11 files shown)


thodson-usgs:
Closing as stale.

The branch is `CONFLICTING`, and it still carries `feat(ogc): add opt-in parallel chunk fan-out` — which landed separately as #341 — so the PR duplicates already-merged history. The `ogc/` package has since been split up in #346 and #348, and is being restructured again in #350/#351, so the files this patches no longer look like this.

The two perf findings are real and are **not** on main yet. Worth redoing against current main:

1. `pd.json_normalize` → `pd.DataFrame` in `_get_resp_data` (now `dataretrieval/ogc/shaping.py:126`). OGC feature properties are flat dicts, so the recursive-nesting path is pure overhead — measured 2.4× faster per page (3.2 ms → 1.35 ms on 2500-feature pages).
2. Skip `drop_duplicates` when the plan has no filter axis (now `dataretrieval/ogc/combining.py:144`). List-axis chunks partition by construction, so the dedup pass can't find anything.

Branch `perf/chunk-client-side` is retained if we want to cherry-pick the perf commit.

PR #368, opened 2026-08-10
refactor(ogc): one entry into the engine; retire fetch_ogc_request

thodson-usgs:
> **Stacked on #367** — this branch includes that PR's commit. Review only the last commit here (`refactor(ogc): one entry into the engine; retire fetch_ogc_request`); merge #367 first, then this PR reduces to its own diff.

## What

Retires `fetch_ogc_request` and gives `get_ogc_data` one extra entry shape — a verbatim `cql_body` — so the OGC facade has a single orchestrated entry that is fully usable through the facade alone.

## Why

`fetch_ogc_request`'s interface could not state its own precondition: it took no `base_url`, so request construction inside it targeted whatever was ambient. Its only production caller, `waterdata.get_cql`, had to import the private ambient to make it usable, and then re-implemented `get_ogc_data`'s body by hand across four OGC internals — the properties id-switch, request construction, execution, and finalization. The deletion test pointed one way: folding that orchestration into the engine concentrates it where id-switching, base-URL handling, and finalization already live.

## How

- `get_ogc_data(..., cql_body=...)` drives the prepared POST as a one-item `FanOut` (ADR 0008's plan shape) with the same retry, progress-line, and interruption semantics as the chunked path — and the same finalizer, so **a resumed `get_cql` now returns the finished `(df, BaseMetadata)` shape**; previously the finalizer was applied outside the executor and `exc.call.resume()` skipped it, returning a raw `(frame, response)` pair. A new test pins this.
- The planner's cql-json passthrough stance is preserved: a verbatim body's size stays the server's judgement (no client-side byte gate on this path).
- `waterdata.cql` consumes only the facade (`get_ogc_data` + `prepare_request_args`), dropping its four private cross-package imports; a new import-linter contract (`cql-facade`) pins that, mirroring the NGWMN facade contract.
- The `ogc` facade shrinks from four names to three; `fetch_ogc_request` is deleted (it was internal — no public-surface change).

## Testing

- Full offline suite: 781 passed (includes the new resume-shape regression test).
- `mypy --strict`, `ruff`, `xenon`, `complexipy`, and `lint-imports` (8 contracts, including the new one) all pass.


thodson-usgs:
## Health metrics vs `main`

*Updated for `8d2eb59e` (the `/simplify` pass) and re-baselined: #367 has merged, so `main` is now `f31b4730` and everything below is **#368 alone** — not the two-PR stack the earlier version of this comment measured.*

Same method as before (pyscn 1.29.0 / radon / xenon / complexipy / import-linter, isolated worktrees, `ProjectRoot` verified).

**pyscn composite: 82/100 (B) — unchanged. Six of seven sub-scores identical to `main`; Architecture drops 87 → 82.**

| | main (`f31b4730`) | this PR (`8d2eb59e`) |
|---|---|---|
| Modules | 56 | 56 |
| Resolved dependency edges | 135 | **132** |
| Max dependency depth | 8 | 8 |
| Technical-debt estimate | 160 h | 160 h |
| Functions | 361 | **359** |
| High-risk entries | 19 | **20** ⚠️ |
| Package LOC (non-blank) | 13,958 | **13,936** |

### Architecture 87 → 82 — new, and it arrived with the `/simplify` commit

This is the one number that genuinely moved, and it is worth being precise about, because it is not what it looks like.

pyscn reports **26 architecture violations on both sides — the same 26 modules, with byte-identical descriptions**. What changes is the severity of exactly two:

```
responsibility  dataretrieval.transport.retry   warning -> error
    "mixes 10 dependency concerns with fan-in 6 and fan-out 5"   (identical text both sides)
responsibility  dataretrieval.waterdata.utils   warning -> error
    "mixes 9 dependency concerns with fan-in 6 and fan-out 5"    (identical text both sides)
```

Weighted violations 26 → 34; compliance 86.73% → 82.38%. Measured across three trees:

| | compliance | banded |
|---|---|---|
| `main` (`f31b4730`) | 86.73% | 87 |
| PR before `/simplify` (`3fcb0242`) | 86.67% | 87 |
| PR head (`8d2eb59e`) | 82.38% | **82** |

I isolated the cause: checking out `8d2eb59e` and reverting **only** `waterdata/cql.py` to its pre-`/simplify` form restores 87, with every other `/simplify` edit still in place.

The trigger is that `cql.py` no longer imports `dataretrieval.ogc` — it now reaches the engine through `waterdata.utils.get_ogc_data`, like every other Water Data getter. That is the change ADR 0007 asks for, and it is what lets the facade-only import contract hold. Removing that edge takes the graph from 135 to 132 resolved edges, and the two modules above tip across a severity band whose own inputs — their fan-in, fan-out, and concern counts — did not change.

I did not reverse-engineer pyscn's severity formula, so I'll claim only what I measured: neither module gained a dependent or a dependency, their violation text is identical on both sides, and the flip is reproducibly caused by removing an edge elsewhere in the graph. That reads as a banding artifact of a *smaller* dependency graph, not as new coupling on `transport.retry` or `waterdata.utils`.

If the banded number matters more than the direction of the change, the way back to 87 is to restore the `cql.py -> dataretrieval.ogc` import — but that reintroduces exactly the duplication the `/simplify` pass removed (`base_url`, `dialect`, `extra_id_cols`, and the `output_id` default restated at the call site). My recommendation is to accept the 82.

### High-risk entries 19 → 20

Still **not** a new complex function — it is `ogc/engine.py`'s *module scope* crossing pyscn's cognitive-complexity-25 line:

```
ogc/engine.py::<module>   cyclomatic 2 -> 2   cognitive 24 -> 26   risk: low -> high
```

The `/simplify` pass shaved this from 27 to 26 by replacing the `_fetch_cql` closure with a `functools.partial`; it is now one point over the threshold.

**18 of the 19 entries already flagged on `main` are `<module>` scopes** (`ogc/planning.py` 140, `nwis.py` 119, `nldi.py` 88, `ogc/requests.py` 83, `progress.py` 82, `transport/fanout.py` 80, …), so at module scope this metric tracks "how much is in the file", not how gnarly any function is.

Per-function on the same change:

- `get_ogc_data` cyclomatic 5 → **6**, cognitive 2 → **3**
- `fetch_ogc_request` and its closure: removed
- no new function at all — `/simplify` replaced the `_fetch_cql` closure with `functools.partial(_walk_pages, GEOPANDAS, row_cap=max_rows)`

`complexipy` — the *per-function* cognitive gate CI actually enforces, also at 25 — passes: the highest function in `engine.py` is `_next_req_url` at **8**, unchanged by this PR, and `get_ogc_data` is **5**.

Splitting the CQL branch into a private `_fetch_cql_body` helper would move engine.py's module aggregate back under pyscn's line. My read is still that it is not worth it — the aggregate would sit just under a threshold it is already only one point over, and the branch reads better where it is, next to the finalizer it shares.

### Elsewhere

- **Maintainability index**: `ogc/engine.py` 61.92 → **63.39** (up — deleting `fetch_ogc_request` more than pays for the added branch). `ogc/requests.py` (49.62) and `ogc/policy.py` (57.20) are unchanged now that #367 is in `main` and no longer part of this diff. `waterdata/cql.py` 84.09 → **75.65** is the one regression; it stays rank A, and MI swings hard on a file this small — it shed ~18 lines while keeping the same two functions.
- **Duplication**: 8.8968% on both sides, with clone groups (5), pairs (50), and total clones (25) **identical**. The denominator effect the earlier version of this comment described is gone now that `main` contains #367.
- **Import contracts**: 8 → **7**, back to `main`'s count — `/simplify` merged the new `cql-facade` contract into `ogc-facade` as a second `source_modules` entry instead of keeping a near-verbatim copy, and updated the ADR 0003 / ADR 0007 compliance paragraphs that described only the NGWMN half of that seam.

Merge gates (xenon `--max-absolute C --max-modules B --max-average A`, complexipy, lint-imports, `mypy --strict`, 781 tests): **pass on both sides**.

PR #387, opened 2026-08-20
Make failures actionable and harden validation, APIs, and CI

thodson-usgs:
## Summary

This PR started as an agent-documentation refresh and an error-message cleanup. Exercising each proposed remedy exposed several cases where following the message caused a second failure or, worse, returned a plausible but wrong result. The scope now covers the resulting validation, correctness, API, geographic-code, test, and CI changes as one caller-correctness pass.

## Actionable argument validation

`dataretrieval._validation` now owns the recurring argument-check shapes and their message vocabulary:

| Helper | Contract |
| --- | --- |
| `require_one_of` | one scalar from a closed vocabulary |
| `require_argument` | one required argument, returned with its type narrowed |
| `require_together` | an all-or-none argument group |
| `require_any_of` | at least one filter from a permissive group |
| `require_exactly_one` | one sufficient alternative, returned as `(name, value)` |
| `reject_together` | mutually exclusive but individually optional arguments |

The helpers are adopted across NLDI, NWDC, NWIS, WQP, OGC dates, Water Data CQL/reference/nearest, and state normalization. Messages identify the caller's parameter spelling, state the problem, and name a correction the caller can execute.

The ratings `file_type` validator remains local deliberately: it validates a collection and reports every invalid member, while `require_one_of` validates one scalar. Their shared formatting is centralized through `render_options`; adding a second collection-helper interface for one caller would be speculative.

### Compatibility note

Malformed argument values now consistently raise `ValueError`. The deprecated `nwis.query_waterdata`, `query_waterservices`, and `get_record` paths therefore no longer use their historic `TypeError` for missing filters, incomplete bounding boxes, or unknown services. `TypeError` remains for genuinely mistyped arguments such as a non-string `sites` value.

## Correctness fixes found by executing the remedies

- NLDI validates origin conflicts in one pass and requires `data_source` before constructing a navigation URL. A missing source can no longer become a literal `/None` path whose 200 response looks like a genuine empty result.
- `waterdata.get_nearest_continuous` adds `time` and `monitoring_location_id` to an explicit `properties` list because both are required to select one row per target and monitoring location. This is a documented returned-column-shape change.
- `nwdc` empty selectors and error envelopes now produce actionable text rather than advice that repeats the same failure or prints a raw validation-object list.
- Shared metadata and state helpers no longer recommend parameters or replacement getters that their caller does not accept.
- NWIS's invalid `query_waterdata("ratings", ...)` route now points to `waterdata.get_ratings`, including the dict return shape and feature-ID keys callers need to use it.
- Date formatting names the public argument, rejects overlong intervals rather than truncating them, and does not advertise durations on the ratings path that refuses them.
- Pagination, credential-install, RDB, and adapter-specific failures now distinguish retryable recovery from deterministic corrections.

`nldi._query_nldi` still turns a 200 non-JSON body into an empty GeoDataFrame. That is an existing documented exception, not changed here, and is called out in `AGENTS.md` so it is not mistaken for the general error policy.

## Additional API and data behavior

- `waterdata.get_cql(..., max_rows=N)` exposes the OGC engine's existing total-row cap. `limit` remains page size; a small `limit` alone can increase request count rather than cap results.
- State normalization now includes American Samoa, Guam, the Northern Mariana Islands, Puerto Rico, and the US Virgin Islands under their ANSI/FIPS codes. Name, postal, FIPS, and `US:<FIPS>` forms normalize alike for adapters whose upstream service supports them.
- The Water Data metadata collection vocabulary includes `countries`, and its hand-maintained reference docstring is contract-tested against that vocabulary.
- `NEWS.md` records the public behavior and exception changes.

## Coverage and CI

Coverage is now a branch-coverage ratchet rather than an informational number. The branch adds behavior-focused tests for dispatch, malformed response envelopes, NLDI navigation, nearest-row grouping, configuration fallbacks, date handling, pagination, and previously uncovered public getters.

The ratchet runs in the existing OS/Python test matrix, blocks on every non-Windows leg, and reports informationally on Windows where POSIX-only skips measure a smaller suite. Version-conditional and environment-unreachable import arms are excluded explicitly so Python 3.10, 3.13, and 3.14 enforce the same measurement. Complexity and dependency-direction jobs remain fast and independent.

## Agent and contributor documentation

- `AGENTS.md` now starts from `CONTEXT.md`, explains placement rules rather than carrying a stale file inventory, names `.importlinter` as the authoritative dependency map, documents the current test/tooling gates, and records the executable-remedy policy.
- `CONTEXT.md` corrects the definition of a setting: a public keyword such as `ssl_check` is not automatically part of configuration resolution.
- `CONTRIBUTING.md` documents branch coverage, platform-dependent local measurements, complexity ratchets, import-contract ownership, and the periodic structural-health sweep.

## Validation

- Focused validation/ratings tests: **45 passed**
- Full offline suite: **1114 passed, 12 deselected**
- Branch coverage: **98.97%**, ratchet met
- `ruff check .` and `ruff format --check .`
- `mypy` in strict mode
- `lint-imports`: **8 contracts kept**
- `xenon` and `complexipy`
- Sphinx HTML build succeeds (5 pre-existing warnings)


.github/workflows/python-package.yml
```diff
@@ -145,13 +145,21 @@ jobs:
...
         # success while tests were red.
         shell: bash
-        run: |
-          coverage run -m pytest tests/
-          coverage report -m
+        run: coverage run -m pytest tests/
+      - name: Coverage ratchet
+        # Windows skips POSIX-only tests, so its number measures a smaller
+        # suite. Negative condition: a matrix edit cannot silently unenforce it.
+        if: runner.os != 'Windows'
```
ehinman, line 159: Aren't these tests run with Linux on GitHub?

dataretrieval/codes/states.py
```diff
@@ -211,11 +223,36 @@ def apply_state(
     native state parameters that must not be combined with ``state``; passing
     ``state`` alongside any of them raises ``ValueError``. Returns the (mutated)
     ``local_vars``.
+
+    An unrecognized ``state`` is re-raised naming the parameters in ``reject``,
+    and only those. They are the endpoint's own state
+    parameters *as the caller spells them*: the mutual-exclusion guard below is
+    proof the getter accepts them as keyword arguments. ``into`` is deliberately
+    not offered, because it is a wire queryable that need not exist on the
```
ehinman: wire queryable?

dataretrieval/codes/states.py
```diff
@@ -211,11 +223,36 @@ def apply_state(
...
+    except ValueError as err:
+        if not reject:
+            # No native spelling of the getter's own to offer instead.
+            raise
+        # Only ``reject`` proves the getter accepts a spelling; ``into`` is the
+        # wire queryable, so it leads only when it appears there too.
+        offered = dict.fromkeys(n for n in (into, *reject) if n in reject)
+        raise ValueError(
+            f"{err} Pass {' or '.join(offered)} instead -- they take the "
+            f"values the API itself uses, unnormalized."
```
ehinman: I don't think this error would make sense to a user, since they wouldn't know what is normalized vs not normalized.

dataretrieval/waterdata/cql.py
```diff
@@ -85,7 +86,16 @@ def get_cql(
         Bounding box ``[xmin, ymin, xmax, ymax]`` in CRS 4326. Combines with the
         CQL filter as an additional spatial predicate.
     limit : int, optional
-        Page size, clamped server-side to 50,000.
+        The number of features returned in each page, clamped server-side to
+        50,000. This is a per-page size, not a cap on the total result: a
+        filter matching more rows than ``limit`` still returns every matching
+        row across multiple pages, so a small ``limit`` makes *more* requests,
+        not fewer. Use ``max_rows`` to cap the total instead.
+    max_rows : int, optional
```
ehinman, line 94: Why is this an input? Trying to think of when someone would want to use this...

dataretrieval/waterdata/nearest.py
```diff
@@ -162,7 +162,11 @@ def get_nearest_continuous(
         Additional keyword arguments forwarded to ``get_continuous``
         (e.g. ``statistic_id``, ``approval_status``, ``properties``).
         Passing ``time``, ``filter``, or ``filter_lang`` raises
-        ``TypeError`` — this function builds those itself.
+        ``TypeError`` — this function builds those itself. A ``properties``
+        list gains ``time`` and ``monitoring_location_id`` if it omits
```
ehinman:
```suggestion
        list gains ``time`` and ``monitoring_location_id`` if the user omits them from the ``properties`` argument
```

dataretrieval/waterdata/nearest.py
```diff
@@ -251,6 +260,25 @@ def get_nearest_continuous(
...
 
 
+def _with_required_properties(kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Keep the columns the nearest-match needs in a caller's ``properties``.
+
+    ``time`` is what the match is computed against and
+    ``monitoring_location_id`` is what it groups by, so a ``properties`` list
+    omitting either silently collapses every site into one row per target
+    rather than failing. Added rather than rejected: the caller asked for
+    columns, not for a lecture about which ones this getter needs.
```
ehinman: This is hilarious.

dataretrieval/waterdata/nearest.py
```diff
@@ -260,8 +288,11 @@ def _select_nearest_rows(
     """Apply the public nearest-per-target shape to continuous rows."""
     if "time" not in df.columns:
         raise ValueError(
```
ehinman: Don't think this is needed anymore, given `_with_required_properties()` adds in these columns if missing.

dataretrieval/ogc/dates.py
```diff
@@ -125,6 +130,17 @@ def _format_api_dates(
     date : bool, optional
         If True, uses only the date portion ("YYYY-MM-DD"). If False (default),
         returns full datetime in UTC ISO 8601 format ("YYYY-MM-DDTHH:MM:SSZ").
+    name : str, optional
+        The caller's own spelling of this argument, used as the subject of
+        every message raised here. Defaults to a generic "date input"; pass
+        the real parameter name (``"time"``, ``"last_modified"``) so a caller
+        correcting the error edits an argument their getter actually accepts.
+    single_value_hint : str, optional
```
ehinman, line 138: This doesn't seem necessary. Is it really that different across functions (?), and the second piece of the error message is a generic set of dates.

dataretrieval/waterdata/ratings.py
```diff
@@ -346,7 +353,12 @@ async def _fetch_rating(
     fid = feature["id"]
     href = _asset_href(feature)
     if not href:
-        raise ValueError(f"STAC feature {fid!r} carries no data asset href.")
+        raise ValueError(
+            f"STAC feature {fid!r} carries no data asset href, so its rating "
+            "cannot be downloaded. Retrying will not help; exclude this "
+            "monitoring location, or report it if the rating is expected to "
+            "exist."
+        )
```
ehinman, line 360: More informative to a human (written with Kiro): "Rating feature {fid!r} has no download URL, so there is nothing "
            "to fetch. Retrying will not help; exclude this monitoring "
            "location, or report it if a rating curve file is expected."

dataretrieval/waterdata/reference.py
```diff
@@ -41,9 +41,11 @@ def get_reference_table(
...
         "aquifer-codes", "aquifer-types", "coordinate-accuracy-codes",
         "coordinate-datum-codes", "coordinate-method-codes", "counties",
-        "hydrologic-unit-codes", "medium-codes", "national-aquifer-codes",
-        "parameter-codes", "reliability-codes", "site-types", "states",
-        "statistic-codes", "topographic-codes", "time-zone-codes"
+        "countries", "hydrologic-unit-codes", "medium-codes",
+        "national-aquifer-codes", "parameter-codes", "reliability-codes",
+        "site-types", "states", "statistic-codes", "topographic-codes",
+        "time-zone-codes". ``METADATA_COLLECTIONS`` is the authoritative
+        list; a test pins this text against it.
```
ehinman: What does this mean?

dataretrieval/_response_metadata.py
```diff
@@ -53,13 +53,16 @@ def __init__(self, response: httpx.Response) -> None:
...
-    # modern ``waterdata`` metadata leaves it unimplemented (use
-    # ``waterdata.get_monitoring_locations`` to retrieve site descriptions).
     @property
     def site_info(self) -> Any:
         raise NotImplementedError(
-            "site_info must be implemented by BaseMetadata children"
+            "This metadata object carries no site_info: only the nwis and wqp "
+            "metadata classes implement it, and the getter that produced this "
+            "result does not return site descriptions alongside data. Fetch "
+            "them from the same adapter -- "
```
ehinman:
```suggestion
            "Fetch them using a separate function -- "
```

dataretrieval/_validation.py
```diff
@@ -1,38 +1,94 @@
...
+who passed a bad ``collection`` that their *code service* was invalid. This
+module owns the wording so a new check cannot invent its own. It is a leaf
+with no first-party imports: the vocabularies live with the adapters that
+define them, and only the rejection is shared.
+
+Three shapes recur: a value outside a closed vocabulary
+(:func:`require_one_of`), a missing argument (:func:`require_argument`,
+:func:`require_together`, :func:`require_any_of`), and arguments that cannot be
+combined (:func:`require_exactly_one`, :func:`reject_together`). Every check
+raises ``ValueError``: each is a complaint about an argument's *value*, and one
```
ehinman:
```suggestion
raises ``ValueError` about an argument's *value*, and one
```

dataretrieval/_validation.py
```diff
@@ -1,38 +1,94 @@
...
-:func:`~dataretrieval.waterdata.get_reference_table` came to tell callers who
-passed a bad ``collection`` that their *code service* was invalid -- the check
-was copied from :mod:`~dataretrieval.waterdata.samples`, message and local
-variable name included, and the noun was never changed.
-
-This module owns the wording so a new check cannot invent its own. It is a
-leaf with no first-party imports: the vocabularies it validates against live
-with the adapters that define them, and only the rejection is shared.
+Rejecting a value outside a closed vocabulary is the one validation every
+adapter does, and it was written eleven times in eight phrasings -- which is
```
ehinman: This seems like Claude is annoyed with what was there before. Is this narrative needed?

dataretrieval/credentials.py
ehinman, line 121: Is this function needed?

dataretrieval/nwdc.py
```diff
@@ -413,7 +405,10 @@ def _read_csv_page(response: httpx.Response) -> pd.DataFrame:
         # zeros, never an empty body — but keep the typed-error contract if it
         # ever returns one rather than leaking a bare pandas exception.
         raise DataRetrievalError(
-            f"NWDC returned an empty response body (URL: {response.url})."
+            f"NWDC returned an empty response body (URL: {response.url}). "
```
ehinman, line 408: Did this happen and that's why there's a guard against it?

dataretrieval/nwis.py
```diff
@@ -37,10 +42,22 @@
...
 WATERSERVICES_SERVICES = ["dv", "iv", "site", "stat"]
+# What ``get_record`` routes, which is wider than what ``query_waterdata``
+# reaches: 'ratings' is served by ``get_ratings`` from a different endpoint.
 WATERDATA_SERVICES = [
     "peaks",
     "ratings",
 ]
+# The major filters each query function accepts, hoisted beside the service
+# lists so the checks and their remedies read from one roster.
+_WATERDATA_MAJOR_FILTERS = ("site_no", "stateCd")
```
ehinman: Why is this named WATERDATA? It's confusing since there's a whole module for water data.

dataretrieval/nwis.py
```diff
@@ -381,24 +402,39 @@ def query_waterdata(
...
+        remedy=(
+            "Pass one, e.g. site_no='01491000' or stateCd='WI', or all four "
+            "bounding-box corners together with "
+            "coordinate_format='decimal_degrees'."
+        ),
+    )
+    require_together(
+        {name: kwargs.get(name) for name in _WATERDATA_BBOX_CORNERS},
+        context="to describe a bounding box",
+        remedy=(
```
ehinman, line 420: Seems like all these remedies could be in a table somewhere rather than hidden in the functions?

.github/workflows/python-package.yml
thodson-usgs: The test matrix is defined a few lines earlier https://github.com/thodson-usgs/dataretrieval-python/blob/5d7ff503220332438cb27b698725926f8c794c3e/.github/workflows/python-package.yml#L128-L135 Which the diff won't show.

ehinman: Duhhh got it, got it.

NEWS.md
```diff
@@ -1,3 +1,7 @@
...
... of `state`, `county` or `huc` must be given, when exactly one was; `codes.states.apply_state` offered NGWMN callers a `state_name` / `state_code` parameter no NGWMN getter accepts; and `BaseMetadata` pointed NGWMN and NWDC callers at a Water Data getter. **Bug fix:** `nldi.get_features(navigation_mode=…)` without a `data_source` spelled `None` into the URL path and returned an empty frame from a 200; it now raises and names the source to pass. **Behavior change:** `waterdata.get_nearest_continuous` appends `time` and `monitoring_location_id` to an explicit `properties` list rather than honoring it verbatim -- omitting either silently collapsed every monitoring location into one row per target, a wrong answer rather than an error -- so a caller who passed `properties=['time', 'value']` now gets a third column. **Behavior change:** `nwis.query_waterdata` serves `'peaks'` only; the `'ratings'` URL it used to build was never an NwisWeb program and answered with an HTML error page. `nwis.get_record(service='ratings')` is unaffected -- it routes to `get_ratings`, which is served from a different endpoint. New: `waterdata.get_cql(..., max_rows=N)` caps the total rows a CQL query returns.
```
ehinman: Suggest using term "user" rather than "caller". This is more relevant to people reading the news.

ehinman (commented):
Interesting work with the error message building. Happy to see additional states and territories added.

ehinman:
Also, it looks like `ogc/retry.py` and `ogc/interruptions.py` aren't needed anymore, since they're just passing through the dataretrieval-level scripts.

dataretrieval/waterdata/nearest.py
thodson-usgs: I think it refers to the properties list here.

dataretrieval/waterdata/reference.py
thodson-usgs: It means there is a unit test that verifies the doc is consistent with the implementation. You could omit that last statement if you like.

dataretrieval/nwis.py
thodson-usgs: I assume it's just follow the precedent from WATERDATA_SERVICES above. I think we can ignore this since the whole module is deprecated.

dataretrieval/waterdata/cql.py
thodson-usgs: `max_rows` was added to prevent agents from blowing my API budget during live tests of chunked queries. During the automated code review, the agent realized it wasn't passed through correctly. Must've broke during an earlier refactor.

dataretrieval/codes/states.py
thodson-usgs: Agreed. ff9294cc replaces “wire queryable” with “API query parameter” in the shared helper and the related Water Data documentation.

dataretrieval/codes/states.py
thodson-usgs: Agreed. The error now says: “Pass state_name or state_code directly instead, using the API’s native value.”

dataretrieval/waterdata/nearest.py
thodson-usgs: I kept this as a response-boundary check rather than removing it. `_with_required_properties()` controls the request, but the service can still omit requested columns. ff9294cc now accepts schema-less empty responses; for nonempty responses it validates both `time` and `monitoring_location_id` and raises `DataRetrievalError` with retry/report guidance.

dataretrieval/ogc/dates.py
thodson-usgs: I kept this customization. The shared formatter accepts durations, but `get_ratings` rejects them; `single_value_hint` prevents its too-many-values message from recommending a duration that the caller’s next attempt would reject.

dataretrieval/waterdata/ratings.py
thodson-usgs: Applied in ff9294cc, including the simpler “no download URL” explanation and rating-curve-file guidance.

dataretrieval/waterdata/reference.py
thodson-usgs: Agreed; ff9294cc removes the test implementation detail from the public docstring.

dataretrieval/_response_metadata.py
thodson-usgs: Applied. It now says to fetch the descriptions using a separate function from the same adapter.

dataretrieval/_validation.py
thodson-usgs: Fixed as part of the module-docstring cleanup in ff9294cc.

dataretrieval/_validation.py
thodson-usgs: Agreed. I trimmed the narrative to the module’s durable responsibilities and invariants: shared validation shapes, `ValueError`, and actionable corrections.

dataretrieval/credentials.py
thodson-usgs: Yes. It centralizes the credential-shaped keyword policy used by both Water Data and WQP, avoiding duplicate lists and preventing plausible credential kwargs from reaching request URLs. I kept it.

dataretrieval/nwdc.py
thodson-usgs: I have not observed the service return an empty body. This is a defensive boundary: it converts pandas `EmptyDataError` into the package’s `DataRetrievalError`; the comment now explicitly says the branch is for an unexpected future response. I kept it.

dataretrieval/nwis.py
thodson-usgs: Agreed after reconsideration. ff9294cc renames the new private constants to `_NWIS_WEB_MAJOR_FILTERS` and `_NWIS_WEB_BBOX_CORNERS` to distinguish the legacy endpoint from the modern `waterdata` module.

dataretrieval/nwis.py
thodson-usgs: I kept the remedies beside the validation calls. The reusable parameter rosters are already hoisted, while each remedy explains its local combination of checks and is not duplicated; a table would separate that explanation from the behavior without adding reuse.

NEWS.md
thodson-usgs: Updated the opening to avoid ambiguity between a human and calling code: “every rejection explains how to correct the call.”

thodson-usgs:
@ehinman Addressed separately in #391: `dataretrieval.ogc.retry` is removed because it exposed only private classifiers, while the published `dataretrieval.ogc.interruptions` path is deprecated for one year before removal so existing imports keep working.

PR #391, opened 2026-08-25
Remove OGC retry and deprecate OGC interruptions

thodson-usgs:
## Summary

Follow-up to review feedback on #387. This removes the obsolete `dataretrieval.ogc.retry` module and deprecates the public `dataretrieval.ogc.interruptions` import path after fan-out execution moved to the shared transport layer.

- `ogc.retry` is removed because its remaining exports were private classifiers. Retry policy lives in `dataretrieval.transport.retry`.
- `ogc.interruptions` remains as an identity-preserving compatibility shim because v1.2.0 exposed public exception classes from that path. Direct imports warn and name `dataretrieval.interruptions` or the top-level exceptions as replacements.
- The deprecated path will be removed in a future major release, no earlier than 2027-08-25.
- Ordinary `import dataretrieval` and `import dataretrieval.ogc` remain silent.
- OGC docstrings now reference the current interruption module, and NEWS contains a concise migration note.

## Validation

- 38 focused contract and architecture tests passed.
- 189 broader affected tests passed.
- Fresh-process import and warning smoke tests passed.
- Ruff lint/format and strict mypy passed.
- Import-linter, Xenon, and complexipy passed.
- All commit hooks passed.


dataretrieval/ogc/interruptions.py
```diff
@@ -1,25 +1,52 @@
-"""Compatibility re-export: the interruption taxonomy moved to a top-level leaf.
-
-The resume contract is no longer OGC-specific — Water Use raises it too — so the
-classes live in :mod:`dataretrieval.interruptions`, where the base class is
-named :class:`~dataretrieval.interruptions.FanOutInterrupted`. This path is kept
-because it is what existing code and tests import; new code should import from
-the leaf, or the top level
-(``from dataretrieval import FanOutInterrupted``).
+"""Deprecated alias for :mod:`dataretrieval.interruptions`.
```
ehinman, line 1: I'm not sure I buy the reasoning behind deprecation vs just removing. Where would any user of this package have loaded ogc interruptions at all?

thodson-usgs: I tend to agree with you. The agent is overly cautious. I was fine to go along, because we'll delete all the deprecated code for the next major release. Will delete now.

thodson-usgs: I reconsidered this. v1.2.0 exposed these exception classes from this module, so deleting it would break an existing public import with no transition. I restored the module as a deprecated alias until a future major release, no earlier than 2027-08-25. `ogc.retry` remains removed because it only exposed private helpers.

ehinman: That's fine.

PR #394, opened 2026-08-26
Simplify WQP module to share internal functions

ehinman:
Drafted by Kiro. This MR allows all WQP functions to leverage the same internal function, _query_wqp, which validates service + profile, makes the API request, detects and formats datetime columns, and creates a metadata object. Reduces redundant code.


thodson-usgs (approved):
The autoreview caught a glossary/adr violation and a minor bug. I went ahead and pushed the fix. Good to merge.

PR #396, opened 2026-08-27
docs: record each explanation once, in the venue that owns it

thodson-usgs:
## Why

Rationale accumulated in docstrings alongside the ADRs that already owned it, because adding a paragraph where the reader is standing is easier than citing a record they would have to open.

An audit of `dataretrieval/` found **~500 lines restating decisions ADRs 0003–0011 already state**:

- `configuration.py`'s module docstring re-derived ADRs 0009/0010/0011 in 65 lines — while citing them three times.
- The no-progress budget was argued from first principles in five places across `transport/retry.py`, `transport/liveness.py`, and `transport/fanout.py`.
- Which failures may be retried was enumerated independently in ADR 0004, ADR 0006, and the architecture overview.

This is a correctness problem rather than a tidiness one. Every copy is a place the rule can be updated without the others, and **two had already gone stale** — an overview paragraph describing concurrency caps a later ADR removed, and an ADR clause describing a credential rejection the code deliberately no longer performs. A reader has no way to tell which copy is current.

## What ADR 0000 decides

Each explanation is recorded once, in the venue that owns that kind of knowledge:

| Kind | Venue |
|---|---|
| The contract — numpydoc `Parameters`, `Returns`, `Raises`, `Examples` | docstring |
| Why *these* lines, when a name cannot carry it | 1–2 line comment |
| Benchmarks, symptoms, what the code used to do | commit message |
| A decision constraining code outside the file stating it | ADR, cited by number |
| A term with package-wide meaning | `CONTEXT.md` |

Three rules follow: **cite rather than restate**, **write from the reader's position**, and **never edit an accepted record to reverse its meaning**.

The middle rule was added after review. The venue rules answer *where* an explanation goes; nothing in them asked *who it reads for*, and review of this PR found eleven places where correctly-placed prose was still unreadable from outside the author's head. One of them broke this record's own history rule — ADR 0012 inventoried the four deprecation spellings module by module — which means the single review question this record shipped with was not enough. `Compliance` now asks a second: *could a reader who has not opened the cited record follow this sentence?*

**Public getter docstrings are explicitly out of scope.** They hold 55% of the package's docstring lines at 2.5 lines of prose per line of code, and that is the deliverable — the policy says so in as many words, so this cannot be read as licence to trim them.

## What changed

- **26 decisions the code was carrying** land as clauses in the records that own their subject (0004, 0006, 0007, 0008, 0009, 0011). New **ADR 0012** records the deprecation policy, which ADR 0005 covered only for legacy NWIS while four hand-rolled shims accumulated.
- **36 sites across 29 files** drop the argument and cite the record — **−189 lines**. Each keeps the sentence saying what the code *does*. No numpydoc section was touched and **no code changed**.
- **Three stale ADR statements marked**: ADR 0004's resume clause (superseded by 0008 without a backlink), the overview's separate concurrency caps (`MAX_CONCURRENT_REQUESTS` is gone), and ADR 0009's claim that the `**queryables` path refuses `session`, which `credentials.py:114` deliberately does not.
- **`CONTEXT.md`** gains `Dialect`, `Single-shot adapter`, `Stall timeout`, `Fitness function` — terms the ADRs used as if defined — and loses two self-contradictions. The largest fix is **facade vs adapter**: the glossary reserves "facade" for a re-export with no logic of its own, while the overview titled the public surface "facades" and then described `nwdc` as one that "builds CSV requests".

## Enforcement

`test_adr_references_resolve_to_a_record` asserts every `ADR NNNN` citation resolves to a record, across **every venue ADR 0000 names** — package, `docs/`, `CONTEXT.md`, `CONTRIBUTING.md`, `AGENTS.md`. A renumbered record fails the suite instead of leaving a dead pointer. It caught one bad citation in this PR's own diff.

Whether a given paragraph *should* have been a citation stays a review judgement. ADR 0000 argues against a proxy metric explicitly: a prose-volume ratio would push contributors to delete parameter documentation to move a number.

## Review notes

- **Amended records say so in `Status`.** Review asked whether these edits should be addenda. They were — `Notes` held the log — but at the bottom, so a reader met the amended Decision text first and the fact of amendment last. `Status` on each amended record now says it was amended and points at `Notes`.
- **Amended records say what changed.** Each notes in `Notes` what was added after acceptance, and where Decision text was *corrected* rather than added to — ADR 0006's retry scope was widened from "gateway 5xx", which was never true of a fanned-out call. No decision is reversed.
- **This PR was reviewed against its own policy** and was breaking it in five places: the semaphore rule went from four copies to five, ADRs 0006 and 0009 each restated the other's credential rules, and four records claimed "the rest is unchanged" while their Decision text had been edited. All fixed before squashing.
- **Removed history is in the commit message**, per the policy — the four-spellings deprecation story, the `base_url` was-once-optional story, and the `_configuration_core.py` benchmarks.

## Conflict with #388

Both branches amend **ADR 0006** and both add a `Notes` section at the end, so they will conflict. It is a straightforward both-added conflict; **land #388 first**, since it is further along.

## Verification

`1130 passed`, `mypy --strict` clean over 60 files, ruff / xenon / complexipy / import-linter all green, every ADR parses clean as RST.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -6,7 +6,7 @@ conversation use them the same way. Architectural decisions live in
 `docs/source/architecture/decisions/`.
 
 When a term here conflicts with a name in the code, the term wins and the name
-is legacy. Legacy names are called out below rather than quietly tolerated.
+is legacy. Legacy names are listed below.
 
 ## Retrieval
 
@@ -47,8 +47,12 @@ A chunk of a large query commonly spans many pages.
 service error, a timeout. Distinguished from a **deterministic failure**, which
 would fail identically every time — an unresolvable hostname, an unsupported
 scheme, a malformed request. Only transient failures are retried, and only
-transient failures produce a resumable interruption. The two answers are one
-judgement about what a failure means, and must agree.
+transient failures produce a resumable interruption. Both answers follow from
+one judgement about what a failure means, and must agree.
+
+**Stall timeout** — How long a call may receive nothing at all before retrying
+stops, measured from when data last arrived rather than from the call's start.
+ADR 0006 sets the policy and calls it the *no-progress budget*.
 
 **Interruption** — A transient failure that stopped a fan-out partway, raised
 with the completed chunks preserved. The caller may wait for the condition to
@@ -112,8 +116,11 @@ elapsed time, and the response headers. Describes the *retrieval*, not the data.
 ## Configuration
 
 **Configuration profile** — A named set of settings for one adapter, stored in
-the configuration file or built in code. **Configuration** is the short form.
-A profile is an *input* to resolution, never its result.
+the configuration file or built in code. A profile is an *input* to resolution,
+never its result. In prose, **configuration** shortens this; in code,
+`Configuration` is the package-wide settings class, and an adapter's profile is
+its own `*Configuration` subclass. The two differ only by case, so prefer the
+full term wherever a reader could take it either way.
 
 **Default profile** — The profile an adapter uses when no other is selected:
 the `[<adapter>]` table's own keys. Always in effect. A **named profile**
@@ -129,8 +136,9 @@ it. **Configure** is the verb for applying one.
 concurrency cap, the retry count, the progress line, the fan-out baseline. A
 setting means the same thing wherever it applies, but it does not apply
 everywhere: `concurrency` is meaningless to an adapter that issues one request
-at a time, and `parallel_chunks` applies only to the two adapters whose queries
-chunk. Which settings an adapter accepts is part of that adapter's vocabulary.
+at a time, and `parallel_chunks` applies only to the adapters whose chunking the
+planner can refine. Which settings an adapter accepts is part of that adapter's
+vocabulary.
 
 A public keyword is not automatically a setting. `ssl_check` is a getter
 argument on four adapters and resolves through no chain at all; the settings are
@@ -145,15 +153,27 @@ alone; it does not replace the package-wide tier. An adapter rejects a setting
 it has no use for, rather than accepting and ignoring it.
 
 The scope is the *adapter*, not the service and not the host, because the
-adapter is what owns the conventions being tuned. The API key is the
-counter-example that fixes the distinction: it belongs to the gateway fronting
-a host, so Water Data and NGWMN — two adapters, one host — necessarily share
-one key and one quota pool. Credentials are host-scoped; tunables are
-adapter-scoped.
+adapter is what owns the conventions being tuned. The API key shows where the
+boundary falls: it belongs to the gateway fronting a host, so Water Data and
+NGWMN — two adapters, one host — necessarily share one key and one quota pool.
+Credentials are host-scoped; tunables are adapter-scoped.
 
-**Source** — Where a setting's value came from. Sources are ordered, and the
-order is resolved per setting rather than per source: a value supplied for one
-setting does not displace another setting's value from a lower source.
+**Source** — Where a setting's value came from, as one of the ordered
+categories: a `configure()` block, the environment, the file, the built-in
+default. The order is resolved per setting rather than per source: a value
+supplied for one setting does not displace another setting's value from a lower
+source. ADR 0010 calls a source a *tier* and ADR 0011 a *rung*; both are this
+term, and the accepted records keep their own wording.
+
+**Origin label** — The exact thing a value came from, at finer grain than its
+source: `$API_USGS_RETRIES`, a path to the config file, the profile a caller
+selected. What `show_configuration()` prints beside each value, and what a
+parser names when it rejects one. A source is the category; an origin label is
+the instance within it.
+
+The code carries both, and spells them the other way around: `_resolve` returns
+its origin label under the name `source` and its source under the name `tier`.
+Prose uses the terms above.
 
 **Selection** — Naming which profile an adapter should use. Done in code; a
 profile is never selected by the environment or implied by the file, so the
@@ -166,8 +186,8 @@ Package-wide.
 supplies one, because that adapter warrants a different figure — NWDC asks for
 4 concurrent requests where the OGC getters take 32. Supplied by the adapter in
 code, not by the user. It replaces the built-in default for calls through that
-adapter and nothing else. A value from any source outranks it: an adapter able
-to override an explicit setting would make that setting a lie.
+adapter and nothing else. A value from any source outranks it: otherwise an
+adapter could discard a value the caller set explicitly.
 
 Distinct from an **adapter-scoped setting**, which is the *user* naming a value
 for one adapter. Both narrow to a single adapter; only one of them is something
@@ -182,6 +202,22 @@ value. Where the distinction matters — reporting what a call will actually use
 **Adapter** — A module owning one service's conventions: its URLs, parameters,
 error shapes, and response quirks. Adapters may use shared machinery; shared
 machinery may not know about adapters.
+
+**Dialect** — The per-API quirks the shared OGC machinery needs in order to
+serve two services from one code path: which collections must be POSTed as
+CQL2, which render dates date-only, which columns to coerce and sort by. An
+adapter supplies one and the machinery reads it, which is how protocol code
+stays free of service names.
+
+**Single-shot adapter** — An adapter whose query is always exactly one request:
+WQP, NLDI, StreamStats, and deprecated NWIS. Nothing divides and nothing
+distributes, so `concurrency` and `parallel_chunks` are not part of its
+vocabulary. NWDC is not one: its query fans out per location even though it
+never chunks by bytes.
+
+**Fitness function** — An executable check that an architectural rule still
+holds, living in `tests/architecture_test.py`. ADR 0003 divides the work between
+these and `.importlinter`.
 
 **Facade** — A module that re-exports a subsystem's public surface and contains
 no logic of its own, so callers depend on a stable name rather than on internal
@@ -207,6 +243,12 @@ Recorded so they are not mistaken for the canonical term, and not re-litigated:
 - `ChunkInterrupted` is a permanent alias of `FanOutInterrupted` — the same
   class object under the name it was first published as. Both spellings are
   correct; neither is scheduled for removal.
+- *No-progress budget* is ADR 0006's name for the **stall timeout**. Both
+  spellings are current; the setting is `stall_timeout`.
+- `ChunkedCall` is a permanent alias of `FanOut`, published on the OGC
+  compatibility path. Like `ChunkInterrupted`, both spellings are correct.
+- `utils.query` is one *request*, not a query as defined above. It is a frozen
+  public path (`dataretrieval.utils.query`) and predates this glossary.
 - `site` appears in deprecated NWIS and WQP parameter names where *monitoring
   location* is meant. These are frozen public surfaces and will not be renamed.
   Where the Water Data API itself names a thing `site-types` or
@@ -216,7 +258,9 @@ Recorded so they are not mistaken for the canonical term, and not re-litigated:
   OGC internals, the Water Data wrappers, and all eleven typed getters now say
   `collection`; `waterdata.get_cql` takes `collection`; and the type alias is
   `WATERDATA_COLLECTIONS`. `service=` on `get_cql` and the `WATERDATA_SERVICES`
-  alias remain — a deprecated keyword and a permanent alias respectively.
+  alias remain — a deprecated keyword and a permanent alias respectively. Two
+  `OgcDialect` fields also still say it: `cql2_services` and
+  `date_only_services` are keyed by collection.
 
   `service` still means the external system in `transport` and `progress`,
   where it labels a progress line. That usage is correct.
@@ -228,8 +272,8 @@ Recorded so they are not mistaken for the canonical term, and not re-litigated:
   query rather than five sets of data, and the OGC definition of *collection*
   is scoped to "access mechanisms defined by OGC API standard(s)", which
   Samples does not implement. Kept as-is by decision: renaming a public
-  keyword costs a deprecation cycle for a term with no better-evidenced
-  replacement in reach.
+  keyword costs a deprecation cycle, and no better-evidenced replacement is in
+  reach.
 - `waterdata.get_codes(code_service=)` is correct and stays. The Samples
   documentation calls it a "code service" in prose and serves it from
   `/codeservice/`, so this reproduces the service's own vocabulary, like
--- a/dataretrieval/_deprecation.py
+++ b/dataretrieval/_deprecation.py
@@ -1,19 +1,10 @@
 """One advisory mechanism, and one place to read the removal horizons.
 
-Four spellings of "tell the caller something is going away" had grown up
-independently -- a dated decorator in :mod:`~dataretrieval.nwis`, an undated
-kwarg shim in :mod:`~dataretrieval.waterdata.utils`, an undated module-level
-notice in :mod:`~dataretrieval.wqp`, and one bare :func:`warnings.warn` with
-no category at all. Only one carried a date, so the horizons could not be
-audited or bumped in one place, and the category was a per-author choice.
-
-The category matters more than the wording. A ``DeprecationWarning`` is a
-promise that a *name in this package* is going away, so a downstream project
-running ``-W error::DeprecationWarning`` is right to fail on it. An advisory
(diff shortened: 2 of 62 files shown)


dataretrieval/ogc/chunking.py
```diff
@@ -174,22 +164,13 @@ def parallel_chunks(n: int) -> Iterator[None]:
...
+    # name ``parallel_chunks(n)`` rather than the ``Configuration`` built
+    # below.
     _configuration._validated_raw("parallel_chunks", n, "parallel_chunks(n)")
-    # Sugar for a package-wide ``Configuration`` rather than a second scope of
-    # its own: two competing ContextVars would let ``show_configuration()`` report a
-    # value the chunker does not use. Sharing one means the innermost block
-    # wins, whichever spelling opened it -- and package-wide rather than scoped
-    # to one adapter, because this block is a per-call request that must reach
-    # whichever adapter the call goes to.
+    # Sugar for the one package-wide ``Configuration`` ContextVar rather than
```
ehinman: What is "sugar" in this context?

dataretrieval/ogc/planning.py
```diff
@@ -621,11 +621,9 @@ def iter_chunk_args(self) -> Iterator[dict[str, Any]]:
                 chunk_args[axis.arg_key] = axis.render(chunk)
             yield chunk_args
 
-    # ``total`` and ``iter_chunk_args`` are this class's domain vocabulary and
-    # stay as they are. The dunders are how a plan satisfies
-    # :class:`~dataretrieval.transport.fanout.FanOutPlan`, which asks for a
-    # sized iterable and nothing chunking-specific. They delegate rather than
-    # duplicate, so ``len(plan)`` cannot disagree with what iterating yields.
+    # The dunders delegate to ``total`` and ``iter_chunk_args`` so a plan
```
ehinman: Dunders?

dataretrieval/_deprecation.py
```diff
@@ -1,19 +1,10 @@
...
-
-The category matters more than the wording. A ``DeprecationWarning`` is a
-promise that a *name in this package* is going away, so a downstream project
-running ``-W error::DeprecationWarning`` is right to fail on it. An advisory
-that an upstream *dataset* has stopped being updated is not that -- the API is
-fine and the caller has nothing to migrate to -- and it belongs under
-:class:`~dataretrieval.exceptions.DataCurrencyWarning` instead. See
-:data:`REMOVALS` for the horizons this package has published.
+Every deprecation is announced through this module, with a horizon in
+:data:`REMOVALS` (ADR 0012). The category is not the author's choice: a
```
ehinman:
```suggestion
:data:`REMOVALS` (ADR 0012). A
```

dataretrieval/configuration.py
```diff
@@ -18,23 +18,17 @@
...
-both name one adapter and two configurations for one adapter raise; source 3
-into the ``[<adapter>]`` table above the top-level keys; and source 4 into an
-adapter's own built-in preference above the package default. That last scope is
-invisible here because this module never supplies it -- it arrives as the
-``default`` a read site like :func:`concurrency` passes for its own service.
+Those are the four *sources*, one branch each in :func:`_resolve`. ADR 0011
+states the same order as seven rungs, splitting three of these sources into the
+scopes inside them. An adapter default is not one of the branches here: it
+arrives as the ``default`` a read site like :func:`concurrency` passes for its
+own adapter.
```
ehinman: No idea what any of this means.

dataretrieval/credentials.py
```diff
@@ -112,7 +101,7 @@ def without_embedded_credentials(url: httpx.URL) -> httpx.URL:
 # substrings without catching legitimate query parameters.
 #
 # ``session`` is deliberately absent from both lists: it carries no secret, so
-# rejecting it with a credentials message told users the wrong thing, and as a
+# rejecting it with a credentials message reported an incorrect reason, and as a
```
ehinman: Weird that this explanation is in past tense. Seems like it should be:
```suggestion
# rejecting it with a credentials message reports an incorrect reason, and as a
```
and continues on below...

dataretrieval/interruptions.py
```diff
@@ -311,9 +310,9 @@ def _classify_transient(
     if isinstance(exc, TransientError):
         return ServiceInterrupted, exc.retry_after
     if isinstance(exc, (httpx.HTTPError, httpx.InvalidURL)):
-        # Some failures will fail the same way every time -- a bad scheme, a
-        # hostname that doesn't resolve. Offering to resume one would just
-        # hide the real error behind a retry that can never work.
+        # Some failures will fail the same way every time -- an unsupported
+        # scheme, a hostname that doesn't resolve. Offering to resume one
+        # would just hide the real error behind a retry that can never work.
```
ehinman:
```suggestion
        # would hide the real error behind a retry that can never work.
```
unnecessary qualifier

docs/source/architecture/decisions/0000-documenting-decisions.rst
```diff
@@ -0,0 +1,127 @@
...
+
+Context
+-------
+
+This package documents itself heavily and deliberately. Its public getters are
+thin wrappers whose numpydoc parameter tables *are* the deliverable: 55% of all
+docstring lines in ``dataretrieval/`` sit in the service adapters, at a ratio of
+2.5 prose lines per line of code. CONTRIBUTING already requires those tables.
+
+The problem is elsewhere. Rationale -- the argument for why a rule holds --
```
ehinman: But it's no longer elsewhere, right? With this PR? Is this being written like it was written before the other ADRs but was actually written after?

docs/source/architecture/decisions/0000-documenting-decisions.rst
```diff
@@ -0,0 +1,127 @@
...
+accumulated in module and function docstrings alongside the ADRs that already
+owned it, because a paragraph can be written where the reader is standing while
+a citation sends them to a record they have to open. Internal modules now hold
+82% of the package's comment lines, and an audit of that prose found roughly 500
+lines restating decisions already recorded in ADRs 0003 through 0011. Three
+examples:
+
+- ``dataretrieval/configuration.py``'s module docstring spends 65 lines
```
ehinman: Are these examples needed?

docs/source/architecture/decisions/0004-error-retry-resume.rst
```diff
@@ -50,4 +88,22 @@ Compliance
...
+-----
+
+The warning, hint-parsing, and pickling clauses were added after the original
+decision. They record, under ADR 0000, rules the code was carrying in prose. The
+skip clause records a carve-out that previously read as contradicting this
+record and :doc:`0006-service-neutral-transport`; 0006 now points here for it.
+
+The ``Status`` line was also annotated retroactively: this record assigned
+resumable partial state to OGC, which :doc:`0008-fan-out-execution` superseded
+without noting it here. The supersession is 0008's; only the backlink is new.
```
ehinman, line 112: Shouldn't these edits be documented as addendums to the original ADR?

docs/source/architecture/decisions/0012-deprecation-horizons.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Context
+-------
+
+This package's value is that established calls keep working. Public API
+compatibility is its first architecture characteristic after artifact integrity.
+Names therefore leave slowly: a renamed argument, a retired module, a getter
+whose service no longer exists.
+
+Four spellings of "tell the caller something is going away" grew up
```
ehinman, line 17: This info isn't needed.

docs/source/architecture/decisions/0012-deprecation-horizons.rst
```diff
@@ -0,0 +1,83 @@
...
+expected to carry one, and an advisory naming a replacement the caller cannot
+yet use is not finished.
+
+``REMOVALS`` is the single table of horizons. One table is auditable -- what is
+due can be listed, and a horizon can be extended in one place -- whereas four
+hand-rolled shims could only be found by grep. A renamed public argument keeps
+working under its old name through one shared decorator rather than a shim
+written for each getter.
+
+The category is not the author's choice. Which category an advisory takes, and
```
ehinman: What category?

docs/source/architecture/index.rst
```diff
@@ -90,25 +90,25 @@ Public service facades
...
+    and protocol modules may depend on it; it must not depend back on them.
+    Scoped overrides use ``ContextVar`` so concurrent threads and asyncio
+    tasks can carry distinct credentials.
 
 ``dataretrieval.ogc``
-    Protocol subsystem for Water Data and NGWMN. A small facade
-    (``__init__.py``) exposes the service-adapter seam: ``OgcDialect``,
-    ``prepare_request_args``, and ``get_ogc_data`` (whose ``cql_body``
-    parameter covers verbatim-CQL2 queries).
+    Protocol subsystem for Water Data and NGWMN. A facade (``__init__.py``)
```
ehinman, line 108: Can we not use "facade"? That's a very AI-y word.

ehinman (approved):
Mostly questions about what the text says and whether it makes sense to add an ADR0000 after ADRs 0001-11 have already been written. Does it make sense to edit the text of an old ADR, or add to it as an addendum? Seemed silly to read through editorial changes to past "Decisions".

dataretrieval/ogc/planning.py
thodson-usgs: "double-underbar" functions

thodson-usgs:
@ehinman, thanks for the review. The idea of AD000 was that it will become a general ADR for documentation that we can reuse across projects. That's the longerterm goal at least. When to create a new ADR versus amend an old one is also TBD. The ADR practice seems to be useful, but we need to adapt it for agents that are less human and more like the guy in Memento.

dataretrieval/ogc/chunking.py
thodson-usgs: Syntactic sugar — a shorthand spelling of something the language already does. Jargon that hides what actually runs, so it is gone: the comment now says the block reuses the one package-wide `Configuration` ContextVar rather than opening a second scope.

dataretrieval/ogc/planning.py
thodson-usgs: Renamed to what they are. The comment now says "`__len__` and `__iter__` delegate to `total` and `iter_chunk_args`", which also tells the reader which two methods are meant.

dataretrieval/_deprecation.py
thodson-usgs: Applied. The clause it drops was commentary on the rule rather than the rule, and the two sentences that follow still say which category goes where.

dataretrieval/configuration.py
thodson-usgs: Fair — the paragraph reconciled the four sources here against ADR 0011's seven rungs, which is unreadable without 0011 open. It now says the two list the same order, that 0011 is the finer grain, and that an adapter default arrives as the `default` argument rather than being a fifth source.

dataretrieval/credentials.py
thodson-usgs: Applied, and carried through the rest of the sentence: "claims part of a namespace the *server* owns" and "would be unreachable". Past tense read as though describing a fixed bug rather than why the list reads as it does.

dataretrieval/interruptions.py
thodson-usgs: Applied.

docs/source/architecture/decisions/0000-documenting-decisions.rst
thodson-usgs: Both readings were fair. "Elsewhere" meant *not in the getter docstrings* — it now says "in the internal modules behind them". And yes: this was written after 0001–0011, numbered 0000 because it governs how every record is written, not because it came first. Notes now says so, and that Context describes what the audit found rather than the state after this PR.

docs/source/architecture/decisions/0000-documenting-decisions.rst
thodson-usgs: Trimmed. The three bullets fold into one sentence in the paragraph above — their file-by-file detail belonged to the audit, not to the decision.

docs/source/architecture/decisions/0004-error-retry-resume.rst
thodson-usgs: They are — `Notes` is the addendum, and no accepted Decision was reversed. The problem was placement: you read the amended Decision text first and learned it had been amended at the bottom. `Status` on 0004, 0006, 0007, 0008, 0009, and 0011 now says the record was amended and points at `Notes`, and ADR 0000 states that as the rule.

docs/source/architecture/decisions/0012-deprecation-horizons.rst
thodson-usgs: Trimmed to the problem. The module-by-module inventory of the four spellings is history the commit that removed them owns — this PR's own policy says so, so keeping it here was inconsistent.

docs/source/architecture/decisions/0012-deprecation-horizons.rst
thodson-usgs: The *warning* category — the paragraph used the word with no antecedent. It now names it, and names both: `DeprecationWarning` for a name in this package going away, `DataCurrencyWarning` for an upstream dataset that stopped updating, with ADR 0004 owning the rule.

docs/source/architecture/index.rst
thodson-usgs: Leaving this one, with the reasoning: "facade" is the Gang of Four pattern name rather than an AI coinage, CONTEXT.md defines it as a re-export with no logic of its own, ADR 0007 is titled with it, and three architecture tests are named for it. Renaming it is a vocabulary change across the ADR set, the glossary, AGENTS.md, and the tests — worth its own PR if you want it. Say the word and I will open one.

ehinman: Fair enough. And I see it defined in the CONTEXT.md In another PR, can we link to the CONTEXT.md in a few more places? Like in the README? This is really helpful for people digging a little deeper into the architecture who are not familiar with the jargon.

thodson-usgs: I recommend something like README -> CONTRIBUTING -> CONTEXT, etc. for humans, and AGENTS.md -> CONTEXT, etc. for bots

CONTEXT.md
```diff
@@ -158,17 +158,9 @@ boundary falls: it belongs to the gateway fronting a host, so Water Data and
...
 order is resolved per setting rather than per source: a value supplied for one
-setting does not displace another setting's value from a lower source. ADR 0010
-calls a source a *tier* and ADR 0011 a *rung*; both are this term, and the
-accepted records keep their own wording.
-
-Distinct from the *label* the code carries alongside it — `API_USGS_RETRIES`,
-or a config-file path — which names the concrete thing a value came from, for
-error messages and `show_configuration()`. The source is the category; the
-label is the instance.
+setting does not displace another setting's value from a lower source.
```
ehinman: This is confusing because you got rid of source in a lot of places and replaced it with tier and don't define tier anywhere.

thodson-usgs: yeah, there's still some glossary drift. I'm hunting for the bug in the agent design, rather than chasing each instance.

thodson-usgs: Correct, and it predates this PR — `"it does not replace the package-wide tier"` was already in CONTEXT.md before this record was written. Recording each explanation once made the gap visible rather than creating it.

Fixing the one word would have left the reason it got there, and this is the third review in a row to land on the same seam, so #405 addresses the seam: the glossary holds two kinds of term. Ones we invented (`chunk`, `page`, `source`) appear in no USGS API and take one spelling everywhere. Ones the services own (`monitoring location`, `collection`) are spelled four different ways upstream — `site_no`, `Station`, `monitoring_location_id` — so the glossary fixes one word for prose while each adapter keeps its own service`'s spelling in its parameters.

That gives your question an answer that does not depend on who reviews: in `nwis` a docstring may say *site*; in `transport` it may not.

thodson-usgs: One correction since: swapping in *source* there was still the wrong axis. Package-wide vs adapter is what ADR 0010 calls a *tier* — its own word for what the glossary now defines as **scope** — while a *source* is block/environment/file/default. The sentence now names neither axis ("leaving that value standing for every other adapter"), and #405 defines **Scope** and the **precedence ladder** so source, scope, and rung each mean one thing.

PR #397, opened 2026-08-29
docs: fix typo concurrancy -> concurrency

vaibhav8a:
Corrects the misspelling `concurrancy` -> `concurrency` in `docs/source/architecture/decisions/0010-adapter-scoped-settings.rst`.

Documentation only, no behaviour change.

### DIFF
--- a/docs/source/architecture/decisions/0010-adapter-scoped-settings.rst
+++ b/docs/source/architecture/decisions/0010-adapter-scoped-settings.rst
@@ -169,14 +169,14 @@ Settings are scoped to the **adapter**, not the service, and not the host.
    with its own ``TypedDict``, so a type checker rejects a setting the adapter
    does not read before the code runs. A ``**unknown`` catch-all remains, and
    exists to turn a misspelled *setting* into a message naming the settings --
-   ``configure(concurrancy=8)`` would otherwise be a bare ``TypeError``.
+   ``configure(concurrency=8)`` would otherwise be a bare ``TypeError``.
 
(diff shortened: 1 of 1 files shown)


thodson-usgs:
Thanks for this, @vaibhav8a — clean, well-scoped, signed off, with an accurate description. I appreciate you taking the time.

I'm going to close it, though, because that particular misspelling is load-bearing. `concurrancy` is the *example* in a paragraph about what happens when you misspell a setting name:

> A ``**unknown`` catch-all remains, and exists to turn a misspelled *setting* into a message naming the settings -- ``configure(concurrancy=8)`` would otherwise be a bare ``TypeError``.

With the correct spelling the sentence no longer demonstrates anything, and the second edit becomes factually wrong — `concurrency` *is* a real field:

```python
>>> WaterdataConfiguration(concurrency=8)    # accepted, no error
>>> WaterdataConfiguration(concurrancy=8)
TypeError: WaterdataConfiguration.__init__() got an unexpected keyword argument 'concurrancy'
```

So the ADR would end up claiming a `TypeError` for a keyword that exists. There's a matching deliberate misspelling in `tests/configuration_test.py::test_a_misspelled_setting_is_not_silently_swallowed`, which fails outright if the same change is applied there.

To be clear, the API itself is spelled `concurrency` everywhere — `configuration.concurrency()`, the `concurrency` field, `API_USGS_CONCURRENT` — so there's no real typo hiding in the code.

Sorry for the wasted effort. Other contributions are very welcome — thanks again for looking.

PR #398, opened 2026-08-30
lint(ruff): select the checks no other gate covers, and pin the linter

thodson-usgs:
Selects the ruff checks that catch something nothing else in the stack can see, fixes what they found, and pins the linter that grades them. **Three live bugs; the suppression budget goes down, not up.**

This started as a pylint adoption and turned into a measurement that pointed the other way — most of pylint's yield was ruff rules that simply were not selected. #399 covers what was left over after that.

## The rules

| Rule | Findings | Why it is not covered elsewhere |
|---|---|---|
| `PLW1514` | 8 (1 in the package) | mypy does not model encodings; a locale-dependent write misbehaves only on the Windows leg, so a green Linux run proves nothing |
| `ARG001` | 5 | the check that notices when retiring a legacy getter (ADR 0005) strands a documented keyword |
| `D100`–`D104` | 1 | `automodule ... :members:` renders a docstring-less public symbol as a bare signature |
| `BLE001` | 0 | already half-adopted — see below |
| `RUF100` | 0 (6 before this PR) | **nothing was watching the suppressions.** Six `# noqa: BLE001` directives sat in the tree suppressing a rule nobody had selected |
| `ASYNC` | 1 | the suite awaits mocked transports, so a stalled event loop still passes |
| `PD` | 4 | `inplace=` is on pandas' way out; this is a pandas library |
| `DTZ` | 1 | a naive `datetime` in a library about time series |
| `C4` `LOG` `G` `T20` `FA` `NPY` | 0 | free ratchets at today's setting |

`BLE001` is the interesting one. Ruff reported all six pre-existing directives as `Unused noqa directive (non-enabled: BLE001)`. Four were on handlers that re-raise and never needed one; those are deleted, and the two on `progress.py`'s best-effort cosmetics now do something. **Four directives removed, one added** (a `DTZ007` that is correct-by-design, see below).

## What they found

- **A silent data loss.** `get_ratings(file_path=...)` wrote each rating with `open(..., "w")` and no encoding. On a non-UTF-8 locale an unrepresentable character raises `UnicodeEncodeError` → a `ValueError` → caught by `_download_all`'s per-feature handler and downgraded to a `SkippedRatingWarning`, so the rating silently disappears from the returned dict. That function's own docstring promises the opposite. It had survived ruff, `mypy --strict`, 995 tests and a 98.9% coverage floor.
- **A stalled event loop.** The same line was a blocking `open()` inside `async def _fetch_rating`, which runs dozens-concurrent under a fan-out drive — so every rating write stalled every other in-flight download for its duration. Now `_write_rating` on a worker thread via `anyio.to_thread.run_sync`.
- **A documented no-op.** `get_record`'s `wide_format`, `datetime_index` and `state` have read nothing since `qwdata`, `gwlevels` and `water_use` were retired, but the docstring still described the behavior they used to have — so `get_record(sites=..., state="OH")` reads as a state filter while doing nothing.
- **A side effect that was already inconsistent.** `format_response` set the index on the caller's frame in place while already returning a new object — and already left the caller's frame untouched on the `peaks` and GeoDataFrame paths, so the in-place branch was the odd one out.

## The flow, not just the rules

The rule list is only half the gate. Four flow defects, none of which selecting rules addresses:

- **The version was pinned in three places and honoured in none.** The pre-commit rev, CI and the `test` extra all said 0.16.1; the machine that wrote this had 0.16.5 on `$PATH` and 0.15.12 in its venv, and `ruff format --check` disagreed by 28 files between them. The pin now lives in **two** places instead of three: a `[lint]` extra that CI installs the way the `[metrics]` job already installs its own, and the `ruff-pre-commit` rev. The `test` extra reuses `dataretrieval[lint]`, exactly as it already reuses `dataretrieval[type-check]`, so a bump is a two-line edit. (A `required-version` guard turned a mismatch into a hard **exit 2**; review pointed out it bought that at the cost of a *fourth* place to keep in step, and it is out.)
- **`preview = true` enabled 52 rules to get one.** `PLW1514` is the only preview rule this config wants; without `explicit-preview-rules` every other preview rule under the selected prefixes is live too — 41 from `E` alone — and each ruff bump adds more. Not hypothetically: that count was **50 at 0.16.1 and is 52 at 0.16.5**, four patch releases later. Those two would have arrived as CI failures nobody selected.
- **Nothing watched the suppressions** — `RUF100`, above.
- **`E501` was selected redundantly** (already inside `E`) under a comment naming a `line-length` setting that did not exist. The value is now set explicitly, where it visibly governs both the checker and the formatter.

A note for authors, now in the config comment: RUF100's fix reads the rest of the line as the directive's description, so a co-located suppression for another tool is deleted with it. Write `# noqa: X # pylint: disable=...`, never `# noqa: X, pylint: ...`. The four directives that carried a reason after a dash or in parens are respelled accordingly.

## Two judgement calls worth your eye

1. **`get_record` warns rather than raises.** My first version raised `NameError`, matching how `get_record` already treats a defunct *service*. Review overturned it, correctly: a defunct service cannot return the data asked for, but `get_record(service="dv", wide_format=False)` returns exactly the right data — only the knob is dead. Raising would retire the parameters ~20 months before `REMOVALS["nwis"]` with no warned release in between. They now warn through `_deprecation.warn_deprecated` and are still ignored.
   They are **not** deleted from the signature: `state` would then fall through `**kwargs` to `query_waterservices` and onto the wire as a parameter NWIS does not define. (`wide_format`/`datetime_index` would not — `to_str` drops non-iterable scalars — so that argument holds for `state` alone.)
2. **ADR 0005 says a deprecated surface should "warn once per user call"** — and a call naming all three options now emits **four** `DeprecationWarning`s: one from `@_deprecated` plus one per option. They are not duplicates (each names a distinct subject and a distinct replacement, which the same ADR sentence asks for), but `_warn_defunct_record_options` raises through `warn_deprecated` directly and so does not pass through `_deprecated`'s re-entrancy guard, which is the module's cadence mechanism. The count is now pinned by a test, as is the attribution the hand-counted `stacklevel=4` encodes.

   Two ways to resolve, your call: clarify the ADR's cadence wording to scope the guarantee per deprecated *surface* (my preference — the code is doing what the Decision explicitly asks), or collapse the three advisories into one warning appended to the function-level message, which costs the per-argument subject/replacement pairing that makes each individually actionable.

## One open question

The ratings write is still text mode (`encoding="utf-8"`, `newline=""`) writing `response.text`. httpx's `TextDecoder` uses `errors="replace"`, so a body that is *not* valid UTF-8 round-trips through U+FFFD — which is normalization, not the byte-for-byte fidelity the test name claims. `open(..., "wb")` + `response.content` would be literally byte-for-byte and needs neither keyword. Practical impact today is nil (USGS RDB is ASCII, and httpx defaults to `utf-8` when the asset carries no charset), and going binary would take the line out of `PLW1514`'s scope, leaving that rule at zero findings in the package. Either is defensible; the write is now isolated in a three-line `_write_rating`, so it is a one-line change. Flagging rather than deciding.

## Cost

Measured at ruff 0.16.1 (the pin at the time), interleaved A/B, n=25, whole tree: **25.9 ms → 25.6 ms**, i.e. −0.3 ms, within noise. For scale, `python -c pass` is 121 ms on the same machine, and pre-commit's own per-hook orchestration is ~380 ms against the ~21 ms of ruff it wraps. The rules cost less than nothing measurable.

## What was rejected, and why it matters

Around 3,500 findings across `A`, `SLF`, `RET`, `DOC`, `FIX`/`TD`, most of pydocstyle and 19 of 20 `PL` codes — none a defect. Several would actively damage the codebase: pydocstyle's section rules ship a **safe** autofix that rewrites eight flagship getters' docstrings into broken ones (they read the indented `Examples:` inside a parameter description as a section header); `DOC502` wants thirteen public getters to delete the `ChunkInterrupted` entry that *is* the resume contract; `PLR0913` fires 40 times on the OGC queryable signatures `CONTEXT.md` names as the public contract. **`PLC0414` must never be selected** — `mypy --strict` implies `no_implicit_reexport`, so the 44 `X as X` aliases in `configuration.py` are required, and removing them produces 29 `[attr-defined]` errors.

Also rejected in the second pass: `N818`/`N801` would rename exported `NWIS_Metadata`, `RateLimited`, `ServiceUnavailable` — a breaking change buying nothing. `S` is 1665 findings (1656 `assert` in tests; the two package hits are correct non-crypto retry jitter). `ANN` is 1722 and `mypy --strict` already covers the package. `PYI019` had one finding but wants `typing_extensions.Self` — a new runtime dependency for one annotation on 3.10.

## Caveats

- `PLW1514` is the only selected rule ruff has not stabilized; it works via `preview = true` and is named explicitly under `explicit-preview-rules`, so a version bump has to re-check it. Re-checked at 0.16.5: still preview-only, still fires.
- `DTZ007` in `ogc/dates.py:_parse_datetime` is correct-by-design — `_DATETIME_FORMATS` deliberately carries both the `%z` and the bare forms, and the docstring promises tz-awareness only for an input that carried an offset. It takes a directive naming the reason rather than a "fix".
- The `tests/*` ignore covers `ARG001` and the five `D1` codes — the largest admission here. pytest matches fixtures and hook arguments by name (renaming `pytest_collection_modifyitems`'s `config` fails collection with a `PluginValidationError`), and test docstrings are not published API. Codes are named rather than written `D1`, so D105–D107 are not pre-exempted. The rationale lives in this description and the commit message rather than in the config, which review asked to keep to the constraint.
- Removing the mypy `anyio` override is unrelated to ruff but was found the same way: its rationale was a 3.9 target that a `match` statement would fail to parse, and `python_version` has been 3.10 since 1.2.0. `mypy --strict` passes without it, in the isolated pre-commit hook env as well.

## Verification

**All 11 CI checks pass**, including the full 6-leg matrix (ubuntu + windows × 3.10/3.13/3.14). `ruff check` + `ruff format --check` clean at the pinned 0.16.5 · `mypy --strict` 60 files clean · **1146 tests pass** · coverage 98.93% against the 98.9 floor · xenon, complexipy, import-linter (8/8 contracts) all pass.

Regression tests added for the fixes. The ratings test asserts the saved file equals the response body byte for byte using a character absent from cp1252, so it fails on the Windows leg without the fix and cannot fail on Linux or macOS.

## Changes after review

- **Ruff bumped 0.16.1 → 0.16.5**, pin and `ruff-pre-commit` rev together. `ruff check` and `ruff format --check` are clean on the tree at that version, and both `PLW1514` and `ARG001` still fire.
- **`required-version` removed** and the `test` extra folded into `dataretrieval[lint]` — two places to keep in step, down from three on `main` and four on the first pass of this branch. The cost is that a contributor on a mismatched local ruff grades differently and quietly, as on `main` today, instead of exiting 2.
- **The ruff config's comments are cut to the constraint** (two or three lines each); the `per-file-ignores` note and the CI step's comment go entirely. The `RUF100` author note about ruff's fix swallowing a co-located `# pylint: disable=` stays — that one is a trap you hit while editing the file, not history.
- **The NEWS entry is two sentences**: the two `waterdata.get_ratings(file_path=...)` bug fixes. The `nwis.get_record` option deprecations and the `format_response` change are no longer called out, `nwis` being deprecated itself.
- `AGENTS.md` now points at the `[lint]` extra rather than at the CI job for the pin.

Still open and unchanged: the ADR 0005 cadence question and the text-versus-binary ratings write below.

https://claude.ai/code/session_01T6MVcko4gh68LieGWxUnrR


.github/workflows/python-package.yml
```diff
@@ -20,8 +20,9 @@ jobs:
           python-version: "3.14"
           cache: "pip"
       - name: Install ruff
-        # Keep this version aligned with the ruff-pre-commit revision.
-        run: pip install ruff==0.16.1
+        # Version pinned in the extra so CI, pre-commit and a contributor's
+        # checkout grade identically; ``required-version`` rejects a mismatch.
```
nodohs:
```suggestion
```

tests/nwis_test.py
```diff
@@ -201,32 +224,92 @@ def test_nested_calls_emit_one_warning(self, httpx_mock):
...
 
-        assert callable(getattr(wd, name, None)), (
-            f"`waterdata.{name}` is missing — fix `_REPLACEMENTS` in nwis.py "
-            "or add the replacement before merging."
-        )
+        The function-level warning is deduplicated by ``_deprecated``'s
+        re-entrancy guard; these advisories are raised directly, so their
+        cadence is pinned here rather than inherited. A call naming all three
+        sees four warnings -- one for the function, one per named option --
+        each with a distinct subject and a distinct replacement.
```
nodohs: I don't think we need to note this was deduplicated. Describe the current implementation.

NEWS.md
```diff
@@ -1,3 +1,7 @@
... locale rather than on what the service sent; on a non-UTF-8 locale an unrepresentable character raised `UnicodeEncodeError`, which is a `ValueError`, which the per-feature handler downgrades to a `SkippedRatingWarning` -- so the rating vanished from the returned dict instead of failing loudly. Ratings are now written as UTF-8 with newline translation disabled, so the saved file is the response body byte for byte on every platform. `dataretrieval.codes` and `dataretrieval.waterdata.types` gained the module docstrings they were missing. **Bug fix:** `waterdata.get_ratings(..., file_path=...)` wrote each rating with a blocking `open()` inside the async fetch, so on a multi-location request every other in-flight download stalled for the duration of each write; the write now runs in a worker thread. **Behavior change:** `nwis.format_response` no longer sets the index on the caller's frame in place. It already returned a new object, and already left the caller's frame untouched on the `peaks` and GeoDataFrame paths, so the in-place branch was the odd one out; callers using the return value -- which is the only documented use -- see no difference. `pandas` is also phasing `inplace=` out.
```
nodohs: this news entry is way to long. including the next paragraph i haven't selected. Please tighten the news to  a few brief sentences describing interface changes and bug fixes to active modules. Don't call out the deprecated nwis changes.

pyproject.toml
```diff
@@ -56,6 +56,13 @@ failed = true
 type-check = [
   "mypy",
 ]
+# The lint gate. Pinned so CI, pre-commit and a contributor's checkout all
+# grade identically; ``required-version`` below turns a mismatch into a hard
+# error rather than a silent difference in findings. Keep the pin, the
+# ``ruff-pre-commit`` rev and ``required-version`` in step.
```
nodohs: omit this comment

pyproject.toml
```diff
@@ -56,6 +56,13 @@ failed = true
...
+lint = [
+  "ruff==0.16.1",
```
nodohs: double check whether we bump this

pyproject.toml
```diff
@@ -111,17 +118,27 @@ repository = "https://github.com/DOI-USGS/dataretrieval-python.git"
 write_to = "dataretrieval/_version.py"
 
 [tool.ruff]
+# A mismatch changes what the gate reports, so fail loudly instead. Keep in
+# step with the [lint] extra and the ``ruff-pre-commit`` rev.
+required-version = "==0.16.1"
```
nodohs: is this necessary? seems like this will be prone to going out of sync.

pyproject.toml
```diff
@@ -111,17 +118,27 @@ repository = "https://github.com/DOI-USGS/dataretrieval-python.git"
...
 target-version = "py310"
+# Ruff's default, set explicitly because it governs both E501 and the
+# formatter -- so the two cannot be given different ideas of a long line.
```
nodohs:
```suggestion
```

pyproject.toml
```diff
@@ -111,17 +118,27 @@ repository = "https://github.com/DOI-USGS/dataretrieval-python.git"
...
+line-length = 88
 extend-exclude = ["demos"]
 
 [tool.ruff.lint]
+# ``preview`` is on for one rule (PLW1514, below). Without
+# ``explicit-preview-rules`` it would also enable every *other* preview rule
+# under the selected prefixes -- 50 of them at 0.16.1, and more on each bump,
+# arriving as CI failures nobody selected. Named preview rules still apply.
```
nodohs:
```suggestion
```

pyproject.toml
```diff
@@ -132,17 +149,69 @@ select = [
...
+    # locale -- cp1252 on the Windows leg -- so a green Linux run proves
+    # nothing. The one selected rule ruff has not stabilized: it works only
+    # under the ``preview = true`` above, and a ruff bump must re-check it.
+    "PLW1514",
+    # ``automodule ... :members:`` ships a docstring-less public symbol to the
+    # docs site as a bare signature. The rest of pydocstyle stays off: its
+    # section rules read the indented ``Examples:`` inside a parameter
+    # description as a section header, and their *safe* autofix rewrites those
+    # correct docstrings into broken ones. ``convention`` is left unset because
+    # setting it would imply more of the family is live than is.
```
nodohs: tighten these comments.

pyproject.toml
```diff
@@ -132,17 +149,69 @@ select = [
...
+    # fan-out getters run dozens of coroutines on one. Nothing else sees it:
+    # the suite awaits mocked transports, so a stalled loop still passes.
+    "ASYNC",
+    # ``inplace=True`` is on pandas' way out, and a naive ``datetime`` in a
+    # library about time series is a bug waiting for a reader in another zone.
+    "PD", "DTZ",
+    # Free ratchets -- zero findings today, each guarding a habit worth
+    # keeping: comprehensions (C4), lazy ``%s`` logging (LOG, G), no ``print``
+    # in a library (T20), ``from __future__ import annotations`` (FA), and
+    # numpy's retired aliases (NPY).
```
nodohs: tighten these comments.

pyproject.toml
```diff
@@ -132,17 +149,69 @@ select = [
...
 [tool.ruff.lint.mccabe]
 max-complexity = 20
 
 [tool.ruff.lint.per-file-ignores]
-"tests/*" = ["SIM108"]
+# A test's parameter names are an interface with something else: pytest matches
+# fixtures and hook arguments by name, so the reflexive ``_x`` rename breaks
+# collection rather than quieting it. Test functions are also not published API
+# -- the reference pages are generated from ``dataretrieval/`` alone. The codes
+# are named rather than written ``D1``, which would also pre-exempt D105-D107.
```
nodohs:
```suggestion
```

nodohs (commented):
please address my suggestions

.github/workflows/python-package.yml
thodson-usgs: Dropped.

pyproject.toml
thodson-usgs: Dropped.

pyproject.toml
thodson-usgs: Dropped. `line-length = 88` stays — it replaces the redundant `E501` entry that was in `select`, and it governs the formatter too.

pyproject.toml
thodson-usgs: Dropped.

pyproject.toml
thodson-usgs: Dropped.

pyproject.toml
thodson-usgs: Cut to two or three lines each — the constraint only, with the reasoning moved to the commit message. Also tightened the `flake8-unused-arguments` note for consistency, since it was the same shape.

pyproject.toml
thodson-usgs: Same — cut to the constraint. The `RUF100` block keeps its author note about ruff's fix eating a co-located `# pylint: disable=`, since that one is a trap you hit while editing the file rather than history.

tests/nwis_test.py
thodson-usgs: Rewritten to describe what the test asserts: a call naming all three options emits four `DeprecationWarning`s — one for `get_record`, one per named option, each with a distinct subject and replacement. The `_deprecated` re-entrancy note is gone.

NEWS.md
thodson-usgs: Cut to two sentences: the two `waterdata.get_ratings(file_path=...)` bug fixes (silent rating loss on a non-UTF-8 locale, and the blocking write stalling the event loop). The `nwis.get_record` option deprecations, the `format_response` in-place change, the module docstrings, and the whole rules paragraph are all out.

pyproject.toml
thodson-usgs: Removed. The sync hazard is real and I had made it worse: the pin was in four places on this branch (`[lint]` extra, `test` extra, `required-version`, `ruff-pre-commit` rev), against three on `main`.

It is now in two — the `[lint]` extra and the pre-commit rev. The `test` extra reuses `dataretrieval[lint]` the way it already reuses `dataretrieval[type-check]`, and CI installs the extra rather than hardcoding a version, so a bump is a two-line edit.

What that costs: a contributor whose local ruff differs now grades differently and quietly, instead of exiting 2. That is the behaviour on `main` today, so nothing regresses — it just stops being caught.

pyproject.toml
thodson-usgs: Bumped to **0.16.5** (the current release), and the `ruff-pre-commit` rev with it. Verified on the tree at that version:

- `ruff check .` — all checks passed; `ruff format --check .` — 98 files already formatted, no reformat.
- `PLW1514` is still preview-only in 0.16.5 (`ruff rule PLW1514` still says the `--preview` flag is required) and still fires on a probe; `ARG001` too.
- Preview rules under the selected prefixes went **50 → 52** between 0.16.1 and 0.16.5 (+2 in `E`/`B`). That drift over four patch releases is exactly what `explicit-preview-rules` is holding back.

Full suite: 1146 passed. `mypy --strict`: 60 files clean.

nodohs (approved):
lgtm

thodson-usgs:
@ehinman , I updated the linter config and this PR fixes bugs and linting errors that resulted from that change. No new features.

thodson-usgs:
Note for whoever merges this, from a separate project rather than a review point on the diff.

`python-maintenance-loop` carries two built-in guardrails — `ruff-selection` and `ruff-config-file` — that refuse *any* automated edit to a ruff configuration file or to a changed `select`/`ignore` line. Their stated reason is this PR series: the rule selection is a surface a human is actively arguing about, so a bot must not touch it.

That reason expires when #398 and #399 land. Worth revisiting the guardrails then rather than leaving a dated justification in place indefinitely — until they are retired, a lint lane there can only report drift and never act on it.

No change requested here.

dataretrieval/waterdata/types.py
```diff
@@ -1,3 +1,18 @@
+"""The Water Data type vocabularies the getters validate their arguments against.
+
+Each ``Literal`` alias here is one closed vocabulary: ``CODE_SERVICES`` for
+the Samples code services, ``METADATA_COLLECTIONS`` for the reference-table
+collections, ``SERVICES`` and ``PROFILES`` for the Samples resources and the
+output profiles they offer, and ``WATERDATA_COLLECTIONS`` (permanent alias
+``WATERDATA_SERVICES``) for the collections ``get_cql`` queries.
```
ehinman: what does this mean? permanent alias?

dataretrieval/nwis.py
```diff
@@ -895,6 +905,41 @@ def what_sites(
...
+# The value that means "not passed" -- which must stay in step with
+# ``get_record``'s declared default -- and the replacement to name.
+_DEFUNCT_RECORD_OPTIONS: dict[str, tuple[object, str]] = {
+    "wide_format": (True, "`waterdata.get_samples()`"),
+    "datetime_index": (
+        True,
+        "`waterdata.get_continuous()` or `waterdata.get_daily()`",
+    ),
+    "state": (None, "`nwdc.get_wateruse(state=...)`"),
+}
```
ehinman, line 919: So this is saying you can't ask for the wide format in the samples function, a datetime index in the water data continuous function, or pass a state in the wateruse function? And these are all passed to `_warn_defunct_record_options()` to let the user know?

dataretrieval/nwis.py
```diff
@@ -1011,6 +1060,10 @@ def get_record(
         ),
     )
 
+    _warn_defunct_record_options(
+        wide_format=wide_format, datetime_index=datetime_index, state=state
```
ehinman, line 1078: Seems weird that you have to do name = name in this function? And it takes as many random things as you give it, right?

tests/nwis_test.py
```diff
@@ -35,6 +39,25 @@
...
 
 
+# Every concrete ``module.function(args)`` the deprecation tables name, so the
+# tripwire below is derived from what ships rather than from a hand-kept list.
+# The prose entries (``waterdata.get_*()``) do not name a function and so do
+# not match.
+_NAMED_REPLACEMENTS = sorted(
+    set(
+        re.findall(
+            r"`(\w+)\.(\w+)\(([^`]*)\)`",
```
ehinman, line 49: Can we explain what this regex means in a comment or in the description of this set?

tests/nwis_test.py
```diff
@@ -200,32 +223,102 @@ def test_nested_calls_emit_one_warning(self, httpx_mock):
         assert len(deprecations) == 1
         assert "get_record" in str(deprecations[0].message)
 
+    @pytest.mark.parametrize("module_name, func_name, arguments", _NAMED_REPLACEMENTS)
+    def test_named_replacement_resolves(self, module_name, func_name, arguments):
+        """Following a deprecation message literally must produce a real call,
```
ehinman: This wording doesn't make sense to me. Is this checking to make sure a deprecation message doesn't error out before the actual call is made?

ehinman (approved):
A few places where the description is insufficient for me to understand, but otherwise pretty straightforward PR. Please consider cleaning up some of the documentation with my comments before merging.

dataretrieval/waterdata/types.py
thodson-usgs: An alias is another name for the same object. Here, `WATERDATA_SERVICES = WATERDATA_COLLECTIONS`, so existing imports of `WATERDATA_SERVICES` still work. “Permanent” means the old name remains supported and is not scheduled for removal; `WATERDATA_COLLECTIONS` uses the collection terminology defined in `CONTEXT.md`. I have prepared a local docstring clarification spelling this out; it is not pushed yet.

dataretrieval/nwis.py
thodson-usgs: The keys are ignored parameters of `nwis.get_record`, not restrictions on the replacement getters. Each tuple contains the default value and the replacement guidance to include in a warning.

- `get_record(wide_format=False)` warns; `waterdata.get_samples` returns one row per result and has no `wide_format` parameter.
- `get_record(datetime_index=False)` warns; `waterdata.get_continuous` and `get_daily` return `time` as a column and have no `datetime_index` parameter.
- `get_record(state="OH")` warns and ignores the state, but `nwdc.get_wateruse(state="OH")` supports that filter.

Yes, `_warn_defunct_record_options()` checks these three values. It emits a parameter warning only when a value differs from its declared default. Local documentation edits clarify the distinction; they are not pushed yet.

dataretrieval/nwis.py
thodson-usgs: This is keyword-argument syntax: in `wide_format=wide_format`, the left side names the argument being passed and the right side reads the local variable. Python does not have a shorthand for this when the names match.

`**given` collects the supplied keyword arguments into a dictionary. The signature accepts any number of keywords, but the body only supports the three names in `_DEFUNCT_RECORD_OPTIONS`; an unknown name would raise `KeyError` at the table lookup. This call passes exactly those three parameters, not the additional `**kwargs` supplied to `get_record`.

tests/nwis_test.py
thodson-usgs: Yes. The regex extracts backtick-enclosed `module.function(arguments)` text from the replacement messages. Its three capture groups return the module name, function name, and argument text. `\w+` matches word characters (letters, digits, and underscores), `\.` and `\(`/`\)` match literal punctuation, and ``[^`]*`` matches text containing no backtick.

For example, the message text `nwdc.get_wateruse(state=...)` produces `("nwdc", "get_wateruse", "state=...")`. The general guidance `waterdata.get_*()` does not match because `*` is not a word character. I have prepared an explanatory comment replacing the “tripwire” wording locally; it is not pushed yet.

tests/nwis_test.py
thodson-usgs: It checks the replacement named in a deprecation message, not the emission of the warning. For each extracted replacement, it verifies that the function exists, is callable, and declares any keyword parameters shown in the message. It never calls that function, so it does not establish that a retrieval succeeds. The original wording overstated what the test checks. I have rewritten the docstring locally to describe these assertions; it is not pushed yet.

thodson-usgs:
The documentation clarifications discussed above are now pushed in fa13fb4e. Merge commit 3197cf91 incorporates main and resolves the WQP test conflict by retaining both explicit UTF-8 encoding and the shared CSV reader. Local verification: 1,177 offline tests passed, coverage 98.93%, Ruff and mypy passed, and all pre-commit hooks passed.

thodson-usgs:
> A few places where the description is insufficient for me to understand, but otherwise pretty straightforward PR. Please consider cleaning up some of the documentation with my comments before merging.

Good feedback, thanks. The agent ran through them and addressed the main documentation gaps. Feel free to open follow up issues if anything is still murky.

PR #418, opened 2026-09-08
docs: copy edits across docstrings, comments, and tests

thodson-usgs:
A second copy editing round. [#417](https://github.com/DOI-USGS/dataretrieval-python/pull/417) covered the ADRs, guides, and glossary; this one extends the same editing to every `.py` file — module and function docstrings, inline comments, and message strings — and makes a further pass over the prose #417 touched, so the source and the documentation use the same words for the same things.

## What changed

Figurative phrasing replaced with the literal statement:

| Instead of | Now |
| --- | --- |
| a dial, a knob, a switch | a setting, a parameter |
| silence, a silent call | time without data |
| a sanctioned wait, a wait the server named | a server-specified wait |
| surfaces as, fires | is raised as, is emitted |
| silently, quietly | without an error, undetected, or dropped |
| lives in, sits in | is defined in, is in |
| drives a fan-out, per drive | runs a fan-out, per run |
| wins, beats, loses to | takes precedence, outranks, is overridden |
| the code knows, wants, asks, answers, trusts | records, requests, responds, accepts |

Emphasis adverbs that carried no fact are dropped: actually, genuinely, precisely, merely, really, and "exactly" where it was not a count. Capitals used for emphasis are lowercased.

## What is unchanged

The glossary's terms stay as `CONTEXT.md` defines them: leaf, facade, seam, fan-out, chunk, page walk, dialect and its quirks, rung and ladder, roster, horizon, no-progress budget with its spend and credit vocabulary, throttle, gate. Standard technical words stay too: driver, backoff, jitter, semaphore, hot path, warm cache, wall clock, drift, stale, leak, in flight. Test function names, fixture data, and quoted upstream text are untouched.

## Notes for review

- The `show_configuration` sample output is duplicated in three places — a docstring, the printed string, and the configuration guide — and `test_show_configuration_sample_output_is_current` compares them. All three were edited together.
- Two message strings changed: the cross-host next-page-link refusal in `transport/links.py`, and the numeric-filter remedy in `ogc/filters.py`. Both are covered by tests.
- The diff is large but mechanical. Reading it by file is probably easier than by hunk.

## Verification

An AST comparison over all 81 changed Python files confirms that only docstrings, comments, and string literals differ — no structural code change. `ruff check`, `ruff format --check`, `mypy --strict`, `lint-imports`, `xenon`, `complexipy`, and the offline suite (1140 passed) all pass.

https://claude.ai/code/session_01N1Ap36zDGnifU8GvjnFowm


thodson-usgs:
@ehinman, this PR consists of one-or-two-word copy edits making the documentation more literal. No code changes

PR #419, opened 2026-09-09
Preserve Samples CSV identifiers through shared parsing

### LINKED ISSUE #411: Preserve Samples identifier columns as text
## What to build

Ensure both Samples CSV getters preserve code and identifier columns as text, including zero-padded parameter codes, hydrologic-unit codes, FIPS codes, and monitoring-location and organization identifiers. Measurement and count columns must retain their appropriate numeric types.

Use one shared response-parsing convention for Samples and WQP rather than importing one adapter from another or duplicating identifier detection. Preserve all existing getter signatures, return shapes, metadata, datetime shaping, and request behavior. Document the intentional dtype correction in the release notes.

## Acceptance criteria

- [ ] Both Samples getters return identifier and code columns as strings without dropping leading zeros.
- [ ] Numeric measurement and count columns remain numeric.
- [ ] Code-column detection and the two-pass CSV parse have one shared implementation used by Samples and WQP.
- [ ] Existing WQP parsing behavior is unchanged and protected by its leading-zero regression coverage.
- [ ] Offline tests cover zero-padded parameter, HUC, and FIPS values through the public Samples getters, including the summary path.
- [ ] Public APIs, request URLs, metadata, frame columns, datetime shaping, and sort behavior are unchanged apart from corrected dtypes.
- [ ] The full offline suite, coverage gate, type checker, linter, architecture tests, and dependency contracts pass.

## Blocked by

None — can start immediately.

tevinch:
Fixes #411.

Both Samples CSV getters currently infer zero-padded codes and identifiers as numbers, so `00060` becomes `60`. Move WQP's two-pass CSV parser into the service-neutral `dataretrieval._csv` module and reuse it from both Samples getters. WQP retains its existing parsing behavior and private helper import.

Samples explicitly leaves `AlternateLocation_IdentifierCount` under numeric inference: despite its name, it is a count, not an identifier. Public getter regressions cover parameter/HUC/FIPS codes, numeric-looking location and organization identifiers, measurement/count dtypes, missing measurements, original date columns, UTC shaping/sorting, URLs, and metadata. Getter signatures and return tuples are unchanged. The dtype correction is documented in NEWS and the Samples docstrings.

Validation on Python 3.12 / pandas 2.2.3:

- `coverage run -m pytest tests/`: 1,142 passed, 12 live tests deselected; coverage 98.93% with branch measurement enabled (98.9% gate).
- `mypy`, `ruff check .`, `ruff format --check .`: passed.
- `lint-imports`: 8 contracts kept; xenon and complexipy passed.
- Built the wheel, verified the shared module is included and matches the source, and imported it outside the checkout.

The two new public-getter cases failed before the identifier fix; the count-field assertions also failed before its inference exception. WQP's regression remains passing.


thodson-usgs:
Thanks @tevinch. I agree, this is the correct behavior but will proceed caustiously since this is a breaking change.

